PR #8, opened 2021-04-23
fs: persist a file record on file creation

yhr:
We're not persisting file records until we get a sync or the
file is closed. This has been observed to cause problems
when recreating the file system state (delete records
of files that does not have an earlier file record in
the meta log)

The above issue has been seen during unclean shutdowns
of rocksdb(reboots, crashes).

Remove the possibility log inconsistency by persisting
a file record when a writable file is created - this
seems to be the right thing to do anyway.


yhr:
Draft for now, i want to complete a full round of tests before merge, pleas review

aravind-wdc (approved):
Looks Good!

yhr:
Regression testing passed

PR #10, opened 2021-04-30
fs: fix parsing of end records

yhr:
End records have no data, just a type, so don't try to read anything
beyond the type.


yhr:
Regression test passed

aravind-wdc (approved):
Looks Good!

PR #11, opened 2021-05-10
test: add zenfs utility related test cases to test suite.

aravind-wdc:
Adds new tests to test zenfs mkfs, list, df, check scheduler,
backup, restore and post-restore data integrity check.

sudo FS_PARAMS="--fs-uri=zenfs://dev:nvme0n1" ./run.sh util-tests utils

### DIFF
--- a/tests/utils/0010_mkfs.sh
+++ b/tests/utils/0010_mkfs.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# Format a ZBD and create a ZenFS filesystem
+
+rm -rf /tmp/zenfs-aux
+$ZENFS_DIR/zenfs mkfs --zbd=$ZDEV --aux-path=/tmp/zenfs-aux --force --finish-threshold=5 > $TEST_OUT
+RES=$?
+if [ $RES -ne 0 ]; then
+  exit $RES
+fi
+
+exit 0
--- a/tests/utils/0030_df.sh
+++ b/tests/utils/0030_df.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# Verify zenfs df command works as expected, simple sanity check.
+
+$ZENFS_DIR/zenfs df --zbd=$ZDEV >> $TEST_OUT
+RES=$?
+if [ $RES -ne 0 ]; then
+  echo "Error: df failed to run" >> $TEST_OUT
+  exit $RES
+fi
+
+exit 0
--- a/tests/utils/common.sh
+++ b/tests/utils/common.sh
@@ -1,0 +1,17 @@
+# Exit on any error
+set -e
+# Helper(s)
+
+utest_run_ldb_verify() {
+  ONDISK_DB="--db=rocksdbtest/dbbench"
+  BACKUP_DB="--db=$1/rocksdbtest/dbbench"
+
+  echo "ONDISK_DB=$ONDISK_DB , BACKUP_DBPATH=$BACKUP_DB, LDB_PARAMS=$LDB_PARAMS" >> $RESULT_DIR/log
+  $TOOLS_DIR/ldb dump --hex $FS_PARAMS $ONDISK_DB > $RESULT_DIR/ondisk_db_dump
+
+  $TOOLS_DIR/ldb dump --hex $BACKUP_DB > $RESULT_DIR/backup_db_dump
+
+  diff $RESULT_DIR/ondisk_db_dump $RESULT_DIR/backup_db_dump
+
+  return $?
+}
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -9,19 +9,21 @@ OK_TESTS=0
 FAILED_TESTS=0
 
 export TOOLS_DIR="../../../"
+export ZENFS_DIR="../util/"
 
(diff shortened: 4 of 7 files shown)


tests/utils/0010_mkfs.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+DEV=${FS_PARAMS:21};
```
yhr, line 3: Hackish. This assumes a lot of the exact fs-uri parameters, so it will easily break. Introduce a new parameter in stead that needs to be set by the user (ZDEV or something like that) running the tests.

sudo FS_PARAMS="--fs-uri=zenfs://dev:nvme0n1" ZDEV=nvme0n1 ./run.sh util-tests utils

tests/utils/common.sh
```diff
@@ -0,0 +1,31 @@
...
+  WORKLOAD=$1
+  DBB_OUT=$2
+  if [ $(grep -wc -E "$WORKLOAD\s+:" $DBB_OUT) -ne 1 ]; then
+    echo "$(tput setaf 1)ERROR: the $WORKLOAD did not complete$(tput sgr 0)" 1>&2
+    return -1
+  fi
+  return 0
+}
+
+utest_run_fillrandom_dbbench_workload() {
```
yhr, line 15: Since this has a single caller, there is no point having it in common.sh, move it in to the caller script in stead, this just makes the code harder to read, same for the others

yhr (changes requested):
Please document what each test is trying to verify in each test case script,  its not that easy to read from the code

We should verify the actual data after backup/restore, not just run readrandom in dbbench as this does not check that we have restored the right data.

See https://github.com/facebook/rocksdb/blob/master/tools/generate_random_db.sh and https://github.com/facebook/rocksdb/blob/master/tools/verify_random_db.sh

Ldb has support for fs-uri, so something similar to those scripts should work.

aravind-wdc:
Thanks for the review. Addressed all previous comments.

tests/utils/common.sh
aravind-wdc: This is a common.sh specific for the utils directory and will have specific functions for these tests.  
This can be used in another script in future if needed, so i thought of keeping it here. I think it will make sense to keep it here for any future needs.

tests/utils/0010_mkfs.sh
aravind-wdc: Ok, ZDEV looks clean.

Yes, you are right, readrandom is more of a sanity test.
So, I have added a check to do a verify after a back up and verify after a restore using "ldb dump" similar to the scripts you pointed out, the steps are as below:
After back up:
Do ldb dump of the ondisk database in to a dumpfileA
Do ldb dump of the backed up directory into dumpfileB
Compare dumpfile A and B, error if it does not match.

After restore:
Do ldb dump of the ondisk database(newly restored) in to a dumpfileA
Do ldb dump of the source directory (from where restore was done) into dumpfileB
Compare dumpfile A and B, error if it does not match.

tests/utils/common.sh
yhr: no, only put things in common.sh if they are common

tests/utils/0030_df.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Verify zenfs df command works as expected.
+
+$ZENFS_DIR/zenfs df --zbd=$ZDEV >> $TEST_OUT
+RES=$?
+if [ $RES -ne 0 ]; then
+  exit $RES
+fi
+
+LINES=$(cat $TEST_OUT | wc -l)
```
yhr, line 11: This just checks that the output is four lines, can't we come up with something that would provide better test coverage?

tests/utils/0030_df.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ $LINES -ne 4 ]; then
```
yhr, line 12: This just checks that the output is four lines, can't we come up with something that would provide better test coverage? Actually sanity checking the numbers?

tests/utils/common.sh
yhr: Put things in common when they are used in more than one place

yhr (changes requested):

tests/utils/0030_df.sh
aravind-wdc: Removed it. Check only if the df subcommand is run successfully.
Also, for the small workload that it runs, the space amp is 0 and does not make sense to check it.

yhr:
Looks good, but please update the commit message to include the ZDEV variable in the example - the mkfs test will fail without it set, right?

aravind-wdc:
> Looks good, but please update the commit message to include the ZDEV variable in the example - the mkfs test will fail without it set, right?

Below usage example is already in commit message.
"sudo FS_PARAMS="--fs-uri=zenfs://dev:nvme0n1" ZDEV=nvme0n1 ./run.sh util-tests utils"

yhr:
Ah, my wrong, the pull request description was not updated, but the commit message was. Pardon! Looks good to me now.

yhr (approved):

PR #12, opened 2021-05-23
Bug fixes

aravind-wdc:


MatiasBjorling:
Hi Aravind, are these good to go?

fs/io_zenfs.h
```diff
@@ -65,6 +66,7 @@ class ZoneFile {
   IOStatus SetWriteLifeTimeHint(Env::WriteLifeTimeHint lifetime);
   std::string GetFilename();
   void Rename(std::string name);
+  time_t GetFileCTime();
```
yhr, line 69: Call this FileCreationTime, no abbreviations please

fs/fs_zenfs.cc
```diff
@@ -574,6 +574,32 @@ IOStatus ZenFS::DeleteFile(const std::string& fname, const IOOptions& options,
   return s;
 }
 
+IOStatus ZenFS::GetFileCreationTime(const std::string& f, const IOOptions& options,
```
yhr, line 577: Implement GetFileModificationTime that is part of the file system api in stead, and use that in the utility

fs/io_zenfs.cc
```diff
@@ -122,6 +126,12 @@ Status ZoneFile::DecodeFrom(Slice* input) {
         extent->zone_->used_capacity_ += extent->length_;
         extents_.push_back(extent);
         break;
+      case kCTime:
+        uint64_t ct;
+        if (!GetFixed64(input, &ct))
+          return Status::Corruption("ZoneFile", "Missing creation time");
```
yhr, line 132: This is breaking backwards compatibility!! so previously created file system cant be mounted!

fs/io_zenfs.cc
```diff
@@ -162,10 +172,13 @@ ZoneFile::ZoneFile(ZonedBlockDevice* zbd, std::string filename,
       fileSize(0),
       filename_(filename),
       file_id_(file_id),
-      nr_synced_extents_(0) {}
+      nr_synced_extents_(0),
+      c_time_(time(0)) {}
```
yhr, line 176: This is changing the storage format of the file system! You need to make the change backward compatible (or step file system version, which does not seem motivated..)

yhr (changes requested):
Please rework the file creation patch(see my comments) and submit a separate pull request for that, it is not a bugfix.. The two other patches look ok

fs/fs_zenfs.cc
aravind-wdc: Ok.

fs/io_zenfs.cc
aravind-wdc: Ok. So if we don't want to bump the filesystem version then I can see 2 options.
     1. Do not error out if c_time_ is not found for a file, report 0(or epoch) for such files.
     2. Hold off this patch until we bump up the filesystem version, then, add a check to not error out if c_time_ is not found and if filesystem's "version_" is older than CURRENT_VERSION.

However, I found this timestamps to be useful for lot of purposes.
Let me know what you prefer.

fs/io_zenfs.h
aravind-wdc: OK

fs/io_zenfs.cc
yhr: I think we can report 0/epoch. And make sure you implement Modification time, not creation time. The RocksDB file system interface only wants modification time, not creation time. Store the creation time and update the value on every sync. But do pull this patch out of this pull request and submit it separately.

aravind-wdc: Ok.
Thanks Hans.

PR #15, opened 2021-06-02
Bug fixes

aravind-wdc:
1. Update the correct file descriptor
2. Update restore path 
3. zenfs currently tries to set the scheduler and needs root privilege. change it to check and report error if mq-deadline is not set.


yhr:
Looks good to me!

PR #16, opened 2021-06-02
Modification time

aravind-wdc:
Add a modification time field to zonefiles.

```
$ sudo ./zenfs list --zbd=nvme0n1 --path=rocksdbtest/dbbench
       56158    Jun 02 2021 09:17:48            LOG
           0    Jun 02 2021 09:17:37            LOCK
   417960339    Jun 02 2021 09:17:42            000012.sst
   421439677    Jun 02 2021 09:17:45            000017.sst
   514117763    Jun 02 2021 09:17:46            000018.log
   242565614    Jun 02 2021 09:17:46            000019.sst
   267304814    Jun 02 2021 09:17:47            000021.log
          16    Jun 02 2021 09:17:37            CURRENT
          37    Jun 02 2021 09:17:37            IDENTITY
         917    Jun 02 2021 09:17:37            MANIFEST-000004
        6143    Jun 02 2021 09:17:37            OPTIONS-000007
```

### DIFF
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -49,6 +49,7 @@ enum ZoneFileTag : uint32_t {
   kFileSize = 3,
   kWriteLifeTimeHint = 4,
   kExtent = 5,
+  kModificationTime = 6,
 };
 
 void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
@@ -72,6 +73,8 @@ void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
     PutLengthPrefixedSlice(output, Slice(extent_str));
   }
 
+  PutFixed32(output, kModificationTime);
+  PutFixed64(output, (uint64_t)m_time_);
   /* We're not encoding active zone and extent start
    * as files will always be read-only after mount */
 }
@@ -122,6 +125,12 @@ Status ZoneFile::DecodeFrom(Slice* input) {
         extent->zone_->used_capacity_ += extent->length_;
         extents_.push_back(extent);
         break;
+      case kModificationTime:
+        uint64_t ct;
+        if (!GetFixed64(input, &ct))
+          return Status::Corruption("ZoneFile", "Missing creation time");
+        m_time_ = (time_t)ct;
+        break;
       default:
         return Status::Corruption("ZoneFile", "Unexpected tag");
     }
@@ -138,6 +147,7 @@ Status ZoneFile::MergeUpdate(ZoneFile* update) {
   Rename(update->GetFilename());
   SetFileSize(update->GetFileSize());
   SetWriteLifeTimeHint(update->GetWriteLifeTimeHint());
+  SetFileModificationTime(update->GetFileModificationTime());
 
   std::vector<ZoneExtent*> update_extents = update->GetExtents();
   for (long unsigned int i = 0; i < update_extents.size(); i++) {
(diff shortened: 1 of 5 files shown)


yhr:
is this ready for review? the file modification time is not set anywhere?

aravind-wdc:
Hi Hans,
Modification time gets updated on the creation of the file, in ZoneFile constructor initialization. Line 180, io_zenfs.cc.
Since sst files are immutable, i thought there would be no modification time to those files and missed the sync case.
But I see that we could change the modification time until the file is being written/synced. And once it is closed, there would not be any changes to modification time, since the file will not change. 
So I will update the modification time in every sync.

Thanks
Aravind

fs/io_zenfs.cc
```diff
@@ -49,6 +49,7 @@ enum ZoneFileTag : uint32_t {
   kFileSize = 3,
   kWriteLifeTimeHint = 4,
   kExtent = 5,
+  kCTime = 6,
```
yhr, line 52: Since we are actually storing modification time, please call this kModificationTime (and no abbreviations please)

fs/io_zenfs.cc
```diff
@@ -83,6 +86,11 @@ Status ZoneFile::DecodeFrom(Slice* input) {
   if (tag != kFileID || !GetFixed64(input, &file_id_))
     return Status::Corruption("ZoneFile", "File ID missing");
 
+  // Zonefiles created by a previous zenfs version may not have the
+  // modification time stamp. Show epoch time stamp if it is not set
+  // previously.
+  m_time_ = 0;
```
yhr, line 92: Good comment, but move m_time_ to the constructor

fs/io_zenfs.cc
```diff
@@ -162,14 +177,16 @@ ZoneFile::ZoneFile(ZonedBlockDevice* zbd, std::string filename,
       fileSize(0),
       filename_(filename),
       file_id_(file_id),
-      nr_synced_extents_(0) {}
+      nr_synced_extents_(0),
+      m_time_(time(0)) {}
```
yhr, line 181: m_time should be 0 after construction(see prev comment), set the modification time when the file is created in stead

yhr (changes requested):
A few small things needs to be addressed(see comments), and drop the non-modification-time commits from this pull request please

fs/io_zenfs.cc
aravind-wdc: Ok, had missed it. Changed now.

fs/io_zenfs.cc
aravind-wdc: Yes, that is a better idea. Now the constructor initializes m_time_ to 0, on file creation, m_time_ is also updated, which will be overwritten on subsequent syncs.

fs/io_zenfs.cc
aravind-wdc: Done.

yhr (approved):

yhr:
Cheers!

MatiasBjorling:
With modification time now different than creation time. Would it make sense to also add creation time to the data structure?

Then we know the time it takes to fill and if any zones takes much longer to fill than others.

yhr:
To figure out the lifetime of a zone(open->finish/reset) we'd need to store metadata about zones, which we do not currently do. We could introduce that in-memory and play around with it to see if we could get some useful metrics from that to start off with.

aravind-wdc:
> Cheers!

Thanks Hans.

PR #18, opened 2021-06-10
fs: use read fd when reading zone report from ZNS device

aravind-wdc:
BLKREPORTZONE ioctl expects the user to have read access,
so use the read fd instead of write fd to get the zone
report from the ZNS device.

Signed-off-by: Niklas Cassel <niklas.cassel@wdc.com>
Signed-off-by: Yura Sorokin <yura.sorokin@percona.com>

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -74,17 +74,16 @@ void Zone::CloseWR() {
 
 IOStatus Zone::Reset() {
   size_t zone_sz = zbd_->GetZoneSize();
-  int fd = zbd_->GetWriteFD();
   unsigned int report = 1;
   struct zbd_zone z;
   int ret;
 
   assert(!IsUsed());
 
-  ret = zbd_reset_zones(fd, start_, zone_sz);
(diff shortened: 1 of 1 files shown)


aravind-wdc:
All zone management IOCTLS requires the user to have root privileges, the following commit from Niklas, is changing that behavior to ensure adequate permissions are there for the user to run zone management commands.
https://lore.kernel.org/linux-block/20210531135444.122018-1-Niklas.Cassel@wdc.com/T/#mceee4bde0f98ab94aa84dbcaa87bf1a9c91420d4

With this above change, BLKREPORTZONE ioctl expects the user to have read permissions on the device.
In Zone::Reset() we pass the write fd to zbd_report_zone(), causing the ioctl to fail, as the write_f_ is not opened with O_RDONLY flag. So pass the read fd to the zbd_report_zones()

We can also wait till the patch is up streamed and available as part of 5.14 kernel.
In current kernels, we cannot run without root privileges, so this change will not have any effect.

yhr:
Looks good, this change makes sense in any case.

yhr (approved):
Looks good

PR #23, opened 2021-07-01
Enable to back up single file or directory

levisonchen:
During our tests on zoned block devices, we need to monitor the database by copying out info logs from ZNS ssd. Original zenfs tool could only backup the whole disk rather than single file or directory, so I add couple few lines of codes to achieve that.

### DIFF
--- a/util/zenfs.cc
+++ b/util/zenfs.cc
@@ -27,6 +27,7 @@ DEFINE_bool(force, false, "Force file system creation.");
 DEFINE_string(path, "", "File path");
 DEFINE_int32(finish_threshold, 0, "Finish used zones if less than x% left");
 DEFINE_string(restore_path, "", "Path to restore files");
+DEFINE_string(backup_path, "", "Path to backup files");
 
 namespace ROCKSDB_NAMESPACE {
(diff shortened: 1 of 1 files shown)


util/zenfs.cc
```diff
@@ -344,7 +345,12 @@ int zenfs_tool_backup() {
     return 1;
   }
 
-  io_status = zenfs_tool_copy_dir(zenFS, "", FileSystem::Default().get(), FLAGS_path);
+  if (!FLAGS_backup_path.empty() && FLAGS_backup_path.back() != '/') {
+    std::string dest_filename = FLAGS_path + "/" + FLAGS_backup_path.substr(FLAGS_backup_path.find_last_of('/')+1);
```
aravind-wdc, line 349: Keep the line to 100 characters max per line.

aravind-wdc:
Thanks for the patch.
Update the commit message with an usage example.
Please follow the commit message format(see previous commits)

levisonchen:
Thanks, I've updated the commit message and changed lines keeping them to 100 characters max.

PR #25, opened 2021-07-06
fs: only acquire lock once on deletion

skyzh:
If there are two threads concurrently calling DeleteFile, it seems possible that there will be a racing condition. This PR fixes this issue.

https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L391

|             Thread 1                |           Thread 2            | |
| ----- | ---- | ---- |
| zoneFile = GetFile(fname); | zoneFile = GetFile(fname); | // both succeed |
|     files_mtx_.lock();           |     waiting                  | |
|   zoneFile = files_[fname]; |     waiting                  |  // succeed |
|    files_mtx_.unlock();        |    files_mtx_.lock();  | |
|                                              | if (zoneFile != nullptr) | // this check passes although zoneFile might have been deleted |
|                                              |   zoneFile = files_[fname]; | // fname is not in files_ any more, and zoneFile will become nullptr |

Point me out if I'm wrong, thanks!

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -378,12 +378,19 @@ IOStatus ZenFS::SyncFileMetadata(ZoneFile* zoneFile) {
   return s;
 }
 
-ZoneFile* ZenFS::GetFile(std::string fname) {
+/* Must hold files_mtx_ */
+ZoneFile* ZenFS::GetFileInternal(std::string fname) {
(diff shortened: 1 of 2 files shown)


aravind-wdc:
This scenario of 2 threads trying to delete same file may not happen in rocksdb. 
However in theory a race can happen in this situation.

PR #32, opened 2021-08-13
util: add zbd micro bench tool

skyzh:
ZBD micro bench tool could help observe how a ZNS device performs when executing different ZBD commands. See README.md for usage information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 util/zenfs
+util/zbd_micro_bench
 fs/*.o
 tests/results
--- a/util/Makefile
+++ b/util/Makefile
@@ -8,10 +8,13 @@ CXX ?= g++
 CPPFLAGS = $(shell pkg-config --cflags rocksdb)
 LIBS = $(shell pkg-config --static --libs rocksdb)
 
-all: $(TARGET)
+all: $(TARGET) zbd_micro_bench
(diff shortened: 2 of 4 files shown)


yhr:
This is out of scope for this project - We have zbd support in fio tool already - see https://zonedstorage.io/benchmarking/fio/ .

yhr (commented):
We use fio for benchmarking zbd devices, see the fio documentation.

MatiasBjorling:
Hi Hans, 

Let's discuss. This tool is useful to validate potential overheads wrt to reset and finish. Which isn't easy to measure with fio at the moment.

yhr:
Well, i don't disagree that might be useful :) a proper, configurable tool
would just do a better job.

AND I don't find any reason why a generic zbd benchmarking  tool should be
added to zenfs. If fio can't be extended, we can see if it would fit in
libzbd or as a separate project.

yhr:
After discussing this with one of our fio developers, it seems feasible to add reset/finish measurements. Having this support in fio would enable a much broader range of different benchmarks. I'll close this if there are no more comments. Will reopen if we get pushback when upstreaming to fio.

skyzh:
Looks good to me, and hope to use this functionality in fio in the future!

PR #37, opened 2021-08-25
Allow zenfs backup to be run on an already mounted ZBDs

percona-ysorokin:
There is no need for 'zenfs' utility to require exclusive access to the drive
when performing backups. Therefore, in the 'backup' command handler 'zbd' is now opened
and 'zenfs' is now mounted in read-only mode.
Please note, that files still being written may be backed up only partially, so it
remains user's responsibility to ensure that the operation they are performing is
safe.

### DIFF
--- a/util/zenfs.cc
+++ b/util/zenfs.cc
@@ -357,10 +357,10 @@ int zenfs_tool_backup() {
   ZonedBlockDevice *zbd;
   ZenFS *zenFS;
 
-  zbd = zbd_open(false);
+  zbd = zbd_open(true);
   if (zbd == nullptr) return 1;
 
-  status = zenfs_mount(zbd, &zenFS, false);
+  status = zenfs_mount(zbd, &zenFS, true);
   if (!status.ok()) {
     fprintf(stderr, "Failed to mount filesystem, error: %s\n",
             status.ToString().c_str());


skyzh (approved):
Looks good so far.

If the disk is been actively written, it is highly possible that metadata is being updated and the tool will report metadata corruption. Maybe we should ignore such corruption when backup at disk writing.

yhr:
@percona-ysorokin : could you point me to the relevant test source code/scripts? (i would like to have a look)  The corruptions from resetting a zone with file data in it is pretty disastrous, and if we can't guarantee that the test is going to be able to cope with that gracefully, we will have a test that will fail randomly for an obscure reason.

percona-ysorokin:
@yhr The MTR test cases I am referring to are
```
rocksdb.compact_deletes
rocksdb.compact_deletes2
```
https://github.com/percona-ysorokin/percona-server/blob/dev/PS-7757-8.0-zenfs/mysql-test/suite/rocksdb/t/compact_deletes.test
and
https://github.com/percona-ysorokin/percona-server/blob/dev/PS-7757-8.0-zenfs/mysql-test/suite/rocksdb/t/compact_deletes2.test
Internally, they are using `sst_count_rows.sh`
https://github.com/percona-ysorokin/percona-server/blob/dev/PS-7757-8.0-zenfs/mysql-test/suite/rocksdb/t/sst_count_rows.sh
where `sst_dump` utility is invoked.

yhr:
Thanks, I'll have a look later this week when I'm back at work again.

percona-ysorokin:
@yhr Thanks, Hans
Just in case, here is how I suggest to rework those 2 MTR test cases provided that you accept the `zenfs backup` readonly fix I submitted.
https://github.com/percona-ysorokin/percona-server/commit/3e22d82332519fd4d8bc0204461aa645071bbe9d

percona-ysorokin:
@yhr Yes, Hans. The tests pass provided that I apply that readonly fix for `zenfs backup`.
`--force` with a scary warning is definitely enough for my tests and will still be safe for end users. So, please go ahead with this option.

yhr:
Resolved in #48

PR #38, opened 2021-08-27
RFC zenfs: do reset and finish operations in a background thread

aravind-wdc:
Due to the nature of finish and reset operations, applications can
experience large latencies when allocating zones to write.
To reduce it, move these operations to a background thread, which
runs whenever a active zone threshold is reached and does the reset
and finish operations in the background.

The allocation logic is same as before and remains as a part of AllocateZone(),
however, it notifies a condition variable on every successful allocation
of an empty zone. The logic to select a zone to reset/finish also remains same,
this patch splits the logic and moves the reset/finish part to a background thread.


yhr:
This work is deprecated by the allocator rework

PR #41, opened 2021-08-29
[WIP] Add `GetZonedBlockDevice` API for fs_zenfs.h

royguo:
Some applications may need to know more details about the devices, e.g. get devices zone stats, file counts etc.

Signed-off-by: Changlong Chen  <levisonchen@live.cn>

### DIFF
--- a/fs/fs_zenfs.h
+++ b/fs/fs_zenfs.h
@@ -362,6 +362,8 @@ class ZenFS : public FileSystemWrapper {
                                 IODebugContext* /*dbg*/) {
     return IOStatus::NotSupported("AreFilesSame is not supported in ZenFS");
   }
+
+  ZonedBlockDevice* GetZonedBlockDevice() { return zbd_; }
 };
 #endif  // !defined(ROCKSDB_LITE) && defined(OS_LINUX)
 


skyzh (commented):
I'm not in favor of this change. See L182

```
  explicit ZenFS(ZonedBlockDevice* zbd, std::shared_ptr<FileSystem> aux_fs,
                 std::shared_ptr<Logger> logger);
```

`ZenFS` is initialized with `ZonedBlockDevice* zbd`, and therefore the one who created `ZenFS` must have a copy of `zbd` itself, and could operate on it freely. There seems to be no need of exporting `zbd` again.

royguo:
> I'm not in favor of this change. See L182
> 
> ```c++
>   explicit ZenFS(ZonedBlockDevice* zbd, std::shared_ptr<FileSystem> aux_fs,
>                  std::shared_ptr<Logger> logger);
> ```
> 
> `ZenFS` is initialized with `ZonedBlockDevice* zbd`, and therefore the one who created `ZenFS` must have a copy of `zbd` itself, and could operate on it freely. There seems to be no need of exporting `zbd` again.


The `zenfs` instance could be passed through the higher-level applications and acts as a context carrier, which will make programming easier for high-level applications. That's why some libraries provide Getters & Setters even their constructors already had the passed-in objects.

IMHO, In the future, `zenfs` should take the role of the context carrier until we defined a new separate `ZenFSContext`.

royguo:
Mark this as a draft since I just find out this PR cannot be merge into master unless the experimental branch is merged first.

PR #46, opened 2021-09-03
fs: refine zone close to prevent potential data racing

royguo:
In `void Zone::CloseWR()`, the `open_for_write_` flag is re-assigned to `false` before it actually finished `Close()`, for the current version this is OK but if we optimize the `AllocateZone` in the future, it will have potential data racing problem.


metaspace:
@royguo Thank you for reporting the potential race, and thank you for the PR! Sorry for the long response time, lots of stuff going on.

This change does not fix the race between `ZonedBlockDevice::AllocateZone` and other `Zone` users.

Closing this PR in favor of #78.

PR #47, opened 2021-09-03
Improve ZonedWritableFile

ls4154:
This patch has follwing changes in `ZonedWritableFile`

`Fsync` is called twice when flush/compaction job finishes sstable, one from sync and the other from close even though there is no file data change between sync and close. I've changed `Fsync` to check last synced position before performing sync.

`PositionedAppend` may not work with buffered option, since `wp` is increased only when buffer is flushed. Although RocksDB does not seems to use `PositionedAppend` in non-direct mode. I changed `wp` to increase when data is appended to buffer.

### DIFF
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -400,6 +400,7 @@ ZonedWritableFile::ZonedWritableFile(ZonedBlockDevice* zbd, bool _buffered,
                                      MetadataWriter* metadata_writer) {
   wp = zoneFile->GetFileSize();
   assert(wp == 0);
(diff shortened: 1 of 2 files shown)


royguo:
For the first issue, maybe check if the `wp` reminds the same after `FlushBuffer` is much simpler? Flushes an empty buffer doesn't hurt anyone and could avoid adding new variables.

ls4154:
What if buffer is empty already? I think we still need to check whether we have new extents since previous fsync.

yhr:
The inline metadata rework has now been merged.

PR #48, opened 2021-09-07
util: allow backups of already mounted filesystems for testing purposes

yhr:
Allow for backing up mounted file systems with a live writer
for testing purposes.

We need some form of snapshots to do this without
risking data corruption/loss, but this is good enough
for now to get zenfs compatible with the mysql/myrocks
mtr tests.

Do this by allowing for non-exclusive block device
access when running the backup command with the --force flag.
A warning is emitted and the the potential data corruption
is documented in the help.

Suggested by Yura Sorokin (in #37 )

Example: (with an active mount by db bench doing writes in another process)
```
$./zenfs backup --zbd=nvme2n1 --path=/tmp/backup_test
Failed to open zoned block device: nvme2n1, error: Invalid argument: Failed to open zoned block device: Device or resource busy
$./zenfs backup --zbd=nvme2n1 --path=/tmp/backup_test --force
Failed to open zoned block device: nvme2n1, error: Invalid argument: Failed to open zoned block device: Device or resource busy
WARNING: attempting to back up a zoned block device in use! Expect data loss and corruption.
rocksdbtest/dbbench/LOG
rocksdbtest/dbbench/LOCK
rocksdbtest/dbbench/000014.log
rocksdbtest/dbbench/000016.sst
rocksdbtest/dbbench/000017.sst
rocksdbtest/dbbench/000018.sst
rocksdbtest/dbbench/CURRENT
rocksdbtest/dbbench/IDENTITY
rocksdbtest/dbbench/MANIFEST-000004
rocksdbtest/dbbench/OPTIONS-000007
```

### DIFF
--- a/util/zenfs.cc
+++ b/util/zenfs.cc
@@ -24,7 +24,11 @@ using GFLAGS_NAMESPACE::SetUsageMessage;
 DEFINE_string(zbd, "", "Path to a zoned block device.");
 DEFINE_string(aux_path, "",
               "Path for auxiliary file storage (log and lock files).");
-DEFINE_bool(force, false, "Force file system creation.");
+DEFINE_bool(force, false, "Force the action. May result in data loss.\n"
+                          "If used with mkfs, data will be lost on an existing "
+                          "file system. If used with backup, data copied from "
+                          "the drive will likely be incomplete and/or corrupt "
+                          "- only use this for testing purposes.");
 DEFINE_string(path, "", "File path");
 DEFINE_int32(finish_threshold, 0, "Finish used zones if less than x% left");
 DEFINE_string(restore_path, "", "Path to restore files");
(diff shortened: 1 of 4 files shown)


yhr:
@percona-ysorokin: looks allright? works for you?

util/zenfs.cc
```diff
@@ -164,7 +168,7 @@ void format_path(std::string &path)
 
 int zenfs_tool_list() {
   Status s;
-  ZonedBlockDevice *zbd = zbd_open(true);
+  ZonedBlockDevice *zbd = zbd_open(true, true);
```
percona-ysorokin, line 171: @yhr I believe here we should not require exclusiveness
```suggestion
  ZonedBlockDevice *zbd = zbd_open(true, false);
```
I don't see any problems with running `zenfs list` on an already mounted ZenFS.
Moreover, `rocksdb.rocksdb_disable_file_deletions` MTR test case depends on this

util/zenfs.cc
```diff
@@ -420,7 +432,7 @@ int zenfs_tool_restore() {
 
 int zenfs_tool_dump() {
   Status s;
-  ZonedBlockDevice *zbd = zbd_open(true);
+  ZonedBlockDevice *zbd = zbd_open(true, true);
```
percona-ysorokin, line 435: Likewise, `zenfs dump` does not need exclusive access.
```suggestion
  ZonedBlockDevice *zbd = zbd_open(true, false);
```

percona-ysorokin (changes requested):

util/zenfs.cc
yhr: Yes, you are right. There is no harm in listing the files, as it is only the integrity of the file data that is at risk when mounting non-exclusive.

yhr:
@percona-ysorokin : Thanks, I incorporated your suggestions and added you as co-author, If you're good and my regression testing passes I'll merge this in later today

percona-ysorokin (approved):
@yhr Thanks Hans
All affected MTR tests are passing now.
It would be really nice if you could merge this to the trunk.

PR #49, opened 2021-09-07
fs: fixed problem with ZenFS::GetChildren() returning error for empty directories

percona-ysorokin:
On ZenFS empty directories cannot be created, therefore we cannot
distinguish between "Directory not found" and "Directory is empty"
and always return empty lists with OK status in both cases.


yhr (changes requested):
Thanks, looks good to me, but can you do a git commit --amend --signoff and prefix the subject with "fs: "?

percona-ysorokin:
@yhr Commit message title changed, `--signoff` added

yhr (approved):

PR #52, opened 2021-09-10
fs: consistent view when zone full

yhr:
AllocateZone might mistakenly assume fully-written zone
is available to use. This is also an issue caused by a race
condition, and will make AllocateZone return nullptr
in rare cases (but actually might happen once a day when
active zone number is close to limit).

A simple fix is to hold the zone_resources_mtx_ throughout
the CloseWR function, instead of taking it twice inside to
NotifyIOZone{Full|Closed}. Therefore, AllocateZone always sees
a consistent state.

(just added more info from Alex to the commit message)

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -65,6 +65,8 @@ void Zone::CloseWR() {
   assert(open_for_write_);
   open_for_write_ = false;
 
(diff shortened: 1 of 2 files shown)


yhr:
Looks like a fine fix to me, let's get this in before we refactor the allocator and improve it.

PR #56, opened 2021-09-16
ci: add zenfs directory as part of lint check.

aravind-wdc:
Only zenfs/fs was being checked earlier,
now all .c/.cc files will be part of the lint check.


aravind-wdc:
I have verified that any errors in both fs/*.cc and util/*.cc were caught by the lint check.

yhr (approved):
looks good to me

Issue #61, opened 2021-09-23
Reduce write latency during metadata roll


yhr:
When rolling to a new metadata zone there is a significant amount write latency introduced.

The write latency is due to holding the file mutex(holding off any other metadata syncs and new file creations etc) while:

1. finishing the current zone
2. writing a superblock and snapshot to new new zone
3. resetting the current metadata zone

We could do the rolling in the background, at the cost of one extra active zone being reserved for metadata rolls.

skyzh:
Metadata zone is mainly written by WAL sync. What about storing WALs in separate zones as logs?

e.g. We can reserve a zone for WAL. Upon WAL sync, we write size + CRC + data to that zone, without touching metadata zone. When we recover from error, we can simply scan the WAL zone.

skyzh:
... and I think this requires a more careful design. What if the "writing a superblock and snapshot to new new zone" is full?

yhr:
@skyzh : A sparse file format is a good idea too. I've actually implemented an extension to the rocksdb WAL format that pads the WAL to be block-size aligned before sync requests (for a different reason - i used the same file format for writing the medata log).

yhr:
If every write is block aligned contains the number of bytes padded, we  don't need to sync metadata and we can recover easily by parsing the file up to the write pointer. We only need to write file metadata when allocating a new extent. This is extra cool because a conventional ssd can't do this :)

yhr:
I decided not to upstream the rocksdb patches as there was no clear performance benefit of it at the time and zenfs would be the only user of it.

yhr:
I think we could make buffered files default to having extent-info being written in-line, and just sync when we need to switch to a new zone (or close the file). I'll dig into the code to see if that is feasible.

royguo:
IIRC, zenfs syncs metadata for SST files only after they were closed, I don’t expect  the inlined sst extents to have more benefits. But inlined WAL metadata looks more general and promising, It only cost us a zone scan on startup which I think is not a big problem.

royguo:
But I should point out that the most significant drawback comes from active zone contention, not the writes itself… so inlined wal metadata is a “better to have”, the first priority to solve this completely is to use a better zone management strategy, LOL

yhr:
[...]

PR #64, opened 2021-10-01
Inline extents

yhr:
Reduce metadata writes for non-direct(buffered) writable files(e.g WAL files) by persisting the data length along with the data when flushing the write buffers. By doing this we only need to do a meta data write sync when switching to a new zone. 

Also allow for recovery of not-yet-synced written data of non-buffered(direct) writable files (e.g. SSTs) using the write pointer of the active extent.

This would minimize the impact of metadata rolls which is a problem if every write is synced. see #61 

This is a breaking/non-backwards compatible on-disk-format change, so this should be merged to master when we step format version next time (so this PR targets the for_v2 branch for now)

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -228,8 +228,8 @@ void ZenFS::LogFiles() {
     ZoneFile* zFile = it->second;
     std::vector<ZoneExtent*> extents = zFile->GetExtents();
 
-    Info(logger_, "    %-45s sz: %lu lh: %d", it->first.c_str(),
-         zFile->GetFileSize(), zFile->GetWriteLifeTimeHint());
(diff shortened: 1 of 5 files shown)


yhr:
Sharing my work so far. 

What is left to do:

- Add a data size prefix for every buffered write
- Only write metadata on close() and when switching to a new zone and on close() (not on every sync) 
- Add a step after mount where the inline extents are recovered from the latest active zone

yhr:
@royguo : Now I'm persisting the data length along with the data for buffered writes.

Todo:
- persist metadata when switching to a new zone 
- get the data back in case of a crash - the recovery is not implemented yet

yhr:
The inline extents implementation has been merged as part of another PR

PR #69, opened 2021-10-27
test: introduce crashtesting

MaisenbacherD:
Adapt crashtesting for ZenFS from the original [db_crashtest.py]
(https://github.com/facebook/rocksdb/blob/master/tools/db_crashtest.py).

It kills the DB at a random point in time (blackbox) or at predefined places
in the RocksDB code (whitebox) and checks for recovery.
For further reading visit the RocksDB
[wiki](https://github.com/facebook/rocksdb/wiki/Stress-test).
However the goal for ZenFS crashtesting is to cover a specified set of
parameters rather than randomized continuous testing.

Therefore db_crashtest.py is copied and modified to use the parameters of
interest. Additionally convenience and integration scripts for the existing
ZenFS test suite are added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,22 @@ for the drive.
 
 `cd tests; ./zenfs_base_performance.sh <zoned block device name>`
 
+
+## Crashtesting
+To run the crashtesting scripts, Python3 is required.
+Crashtesting is done through the modified db_crashtest.py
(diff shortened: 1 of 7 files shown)


yhr:
Looks good, i'll try it out before merging it.

yhr (approved):
Looks like a great start

Issue #68, opened 2021-10-27
Failure in Zone::CloseWR is silently ignored


metaspace:
If a zone fails to close, the error is silently ignored. This will cause ZenFS to eventually deadlock. The error should propagate and cause ZenFS to abort, or at least log an error message.

PR #70, opened 2021-10-27
Do not silently ignore zone close errors

metaspace:
Suggestion for fix fix for #68

### DIFF
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -222,7 +222,7 @@ ZoneFile::~ZoneFile() {
   }
   IOStatus s = CloseWR();
   if (!s.ok()) {
-    abort();
+    zbd_->SetZoneDeferredStatus(s);
   }
 }
 
@@ -476,7 +476,7 @@ ZonedWritableFile::~ZonedWritableFile() {
   if (buffered) free(buffer);
 
   if (!s.ok()) {
-    abort();
+    zoneFile_->GetZbd()->SetZoneDeferredStatus(s);
   }
 }
 
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -220,18 +220,25 @@ ZoneFile::~ZoneFile() {
     zone->used_capacity_ -= (*e)->length_;
     delete *e;
   }
-  CloseWR();
+  IOStatus s = CloseWR();
+  if (!s.ok()) {
+    abort();
+  }
 }
 
-void ZoneFile::CloseWR() {
+IOStatus ZoneFile::CloseWR() {
+  IOStatus s = IOStatus::OK();
+
   if (active_zone_) {
-    IOStatus status = active_zone_->CloseWR();
+    s = active_zone_->CloseWR();
     ReleaseActiveZone();
-    if (status.ok()) {
+    if (s.ok()) {
       zbd_->NotifyIOZoneClosed();
     }
   }
   open_for_wr_ = false;
+
+  return s;
 }
 
 void ZoneFile::OpenWR() { open_for_wr_ = true; }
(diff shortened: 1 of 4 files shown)


metaspace:
@yhr Added a flag that is checked during zone allocation. If any unreported errors were posted, the error is reported by next call to AllocateZone.

fs/zbd_zenfs.h
```diff
@@ -79,6 +81,8 @@ class ZonedBlockDevice {
   std::atomic<long> active_io_zones_;
   std::atomic<long> open_io_zones_;
   std::condition_variable zone_resources_;
+  std::mutex zone_close_mutex_;
+  boost::optional<IOStatus> zone_close_error_;
```
yhr, line 85: Why do we need optional here? why not just set it to ok?

fs/io_zenfs.cc
```diff
@@ -220,15 +220,22 @@ ZoneFile::~ZoneFile() {
     zone->used_capacity_ -= (*e)->length_;
     delete *e;
   }
-  CloseWR();
+  IOStatus status = CloseWR();
```
yhr, line 223: We use "s" for status everywhere else, so please use it here as well (and in the other places you add io status variables)

fs/zbd_zenfs.cc
```diff
@@ -467,13 +477,23 @@ void ZonedBlockDevice::ResetUnusedIOZones() {
   }
 }
 
-Zone *ZonedBlockDevice::AllocateZone(Env::WriteLifeTimeHint file_lifetime) {
+IOStatus ZonedBlockDevice::AllocateZone(Env::WriteLifeTimeHint file_lifetime,
+                                        Zone **out_zone) {
```
yhr, line 481: "out_zone" -> allocated_zone

fs/zbd_zenfs.h
yhr, line 85: Let's make this "zone_close_error_" a generic "zone_management_status" in stead so we can use it for other stuff that goes wrong in the future. zone_management_status should be initialized to ok.

fs/zbd_zenfs.cc
```diff
@@ -467,13 +477,23 @@ void ZonedBlockDevice::ResetUnusedIOZones() {
...
   Zone *finish_victim = nullptr;
   unsigned int best_diff = LIFETIME_DIFF_NOT_GOOD;
   int new_zone = 0;
   Status s;
 
+  *out_zone = nullptr;
+
+  {
+    std::lock_guard<std::mutex> lock(this->zone_close_mutex_);
+    if (this->zone_close_error_ != boost::none) {
```
yhr, line 492: you don't need this->, and you can just use if(!zone_close_error->ok()) which is a lot easier to read and understand

yhr (changes requested):

fs/zbd_zenfs.h
metaspace: I think the intention of the code carries more clearly when using an option.

fs/io_zenfs.cc
metaspace: Just trying to introduce some best practice. Using descriptive names for variables carries intent more clearly for the reader.

fs/zbd_zenfs.cc
metaspace: Prefixing pointer arguments used for returning allocated items with `out_` conveys intention to the reader. How about `out_allocated_zone` ?

fs/zbd_zenfs.cc
metaspace: Using the `this` pointer when accessing class members makes the code easier to read, as it is easier to see if a stack local variable or a class member that is referenced.

yhr: The underscore indicates that it is a class member, and we don't use this-> anywhere else (and we're not converting the rest of the code).  / code style dictator

fs/zbd_zenfs.cc
yhr: that would be fine

fs/io_zenfs.cc
yhr: s is fine and used everywhere else

fs/zbd_zenfs.h
yhr: We want the status to be ok per default

metaspace:
@yhr rebased on master and addressed your comments :)

metaspace:
Changed commit messages to match style.

yhr (approved):

PR #71, opened 2021-10-28
fix: Prevent data race when closing writable zone

metaspace:
This patch prevents a race condition on Zone::open_for_write_ by locking on
access to the field.

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -60,11 +60,21 @@ bool Zone::IsFull() { return (capacity_ == 0); }
 bool Zone::IsEmpty() { return (wp_ == start_); }
 uint64_t Zone::GetZoneNr() { return start_ / zbd_->GetZoneSize(); }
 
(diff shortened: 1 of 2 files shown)


metaspace:
Made a version with more coarse locking: https://github.com/metaspace/zenfs/tree/zone-locking-coarse

yhr (changes requested):
The zone resources mtx is protecting the active / open counters, and I don't think it is a good idea to extend its scope.

Note that zone->IsUsed() also checks if the zone is open for writing.

I think what we want going forward is a "Busy flag" per zone, that marks the zone as having an active writer, is being reset or being finished. Then we can handle all of these things concurrently in a safe way, and we can get rid of the big io zone mutex eventually.

We would then need  zone->SetBusy(bool) (that would return if we managed to transit the zone to busy/not busy)and zone->IsBusy(bool)

SetBusy/IsBusy would replace the open-coded open_for_write_ accesses in the code

metaspace:
Closed in favor of #78

PR #74, opened 2021-11-01
feat: Add runtime check for BLKCLOSEZONE ioctl

### LINKED ISSUE #66: support for BLKCLOSEZONE should be required
The latest Ubuntu 20.04.3 LTS has kernel 5.11 but only linux-libc-dev version (5.4.0-89.100) what means that blkzoned.h is from 5.4 and doesn't support BLKCLOSEZONE.
IMHO it doesn't make sense to compile `zenfs` without BLKCLOSEZONE because it doesn't work correctly. It would be better to stop configure or building with an error message that blkzoned.h from 5.5+ is required.

@damien-lemoal wrote:
Zenfs can trivially check for that on startup by trying to open+close a full or empty zone (that is a no-op normally). If zenfs sees an ENOSUP error from libzbd, then a badly compiled version of libzbd is being used and zenfs can warn the users. Again, nothing wrong with libzbd here. zenfs should handle this issue, and a build problem reported to Ubuntu. Distros normally rebuild all packages that have an uapi dependency against newer kernels. The libzbd .deb packaging definitions may have a problem (that is not maintained in libzbd project, eventhough that is synced from time to time).

metaspace:
ZenFS requires a kernel with support for BLKCLOSEZONE ioctl. This patch adds a
check for this ioctl when a zoned block device is opened by ZenFS.

Closes #66

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -4,9 +4,8 @@
 //  COPYING file in the root directory) and Apache 2.0 License
 //  (found in the LICENSE.Apache file in the root directory).
 
+#include <cerrno>
 #if !defined(ROCKSDB_LITE) && !defined(OS_WIN)
-
-#include "zbd_zenfs.h"
 
 #include <assert.h>
 #include <errno.h>
@@ -25,6 +24,7 @@
 #include <vector>
 
 #include "rocksdb/env.h"
+#include "zbd_zenfs.h"
 
 #define KB (1024)
 #define MB (1024 * KB)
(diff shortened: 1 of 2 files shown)


metaspace:
BLKCLOSEZONE was introduced in linux-5.5. NVMe-ZNS was added in linux-5.9. So one would only hit this if running against `null_blk` or SMR before linux-5.5. If you run against a ZNS device before linux-5.9, you get:
```
Failed to open zoned block device: nvme0n1, error: Invalid argument: Failed to open zoned block device: No such file or directory
```
While (with this patch) on nullb or SMR before linux-5.5 you get:
```
Failed to open zoned block device: nullb0, error: Not implemented: IOCTL BLKOPENZONE not supported
```

Not merging this, since it is not relevant because of above analysis.

Issue #51, opened 2021-09-09
Memory corruption when DeleteFile() is called for a file explicitly closed by Close()


percona-ysorokin:
`rocksdb.use_direct_io_for_flush_and_compaction` MTR test case run under ASan/Valgrind helped to reveal the following `heap-use-after-free` problem.
```
==527273== Invalid read of size 8
==527273==    at 0x2C039C94: rocksdb::ZoneFile::CloseWR() (io_zenfs.cc:225)
==527273==    by 0x2C03A9D3: rocksdb::ZonedWritableFile::~ZonedWritableFile() (io_zenfs.cc:424)
==527273==    by 0x2C03AA1B: rocksdb::ZonedWritableFile::~ZonedWritableFile() (io_zenfs.cc:426)
==527273==    by 0x2B9575EB: std::default_delete<rocksdb::FSWritableFile>::operator()(rocksdb::FSWritableFile*) const (unique_ptr.h:85)
==527273==    by 0x2B9562F7: std::unique_ptr<rocksdb::FSWritableFile, std::default_delete<rocksdb::FSWritableFile> >::~unique_ptr() (unique_ptr.h:361)
==527273==    by 0x2BC7695D: rocksdb::(anonymous namespace)::CompositeWritableFileWrapper::~CompositeWritableFileWrapper() (composite_env.cc:105)
==527273==    by 0x2BC76989: rocksdb::(anonymous namespace)::CompositeWritableFileWrapper::~CompositeWritableFileWrapper() (composite_env.cc:105)
==527273==    by 0x2BB116CF: std::default_delete<rocksdb::WritableFile>::operator()(rocksdb::WritableFile*) const (unique_ptr.h:85)
==527273==    by 0x2BB1140B: std::unique_ptr<rocksdb::WritableFile, std::default_delete<rocksdb::WritableFile> >::~unique_ptr() (unique_ptr.h:361)
==527273==    by 0x2C04F0F4: myrocks::rocksdb_init_internal(void*) (ha_rocksdb.cc:5582)
==527273==    by 0x2C05336C: myrocks::rocksdb_init_func(void*) (ha_rocksdb.cc:6144)
==527273==    by 0x3A9FFF0: ha_initialize_handlerton(st_plugin_int*) (handler.cc:801)
==527273==    by 0x3FA06AA: plugin_initialize(st_plugin_int*) (sql_plugin.cc:1288)
[...]

PR #76, opened 2021-11-03
fs: ensure ZoneFile object is reference counted

aravind-wdc:
Maintain the zonefile objects as a shared pointer to ensure the objects
are not accidentally deleted when it is still held by other users.

Refer issue #51 for details.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -220,12 +220,12 @@ ZenFS::~ZenFS() {
 }
 
 void ZenFS::LogFiles() {
-  std::map<std::string, ZoneFile*>::iterator it;
+  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   uint64_t total_size = 0;
 
   Info(logger_, "  Files:\n");
   for (it = files_.begin(); it != files_.end(); it++) {
-    ZoneFile* zFile = it->second;
+    std::shared_ptr<ZoneFile> zFile = it->second;
     std::vector<ZoneExtent*> extents = zFile->GetExtents();
 
     Info(logger_, "    %-45s sz: %lu lh: %d", it->first.c_str(),
@@ -245,9 +245,9 @@ void ZenFS::LogFiles() {
 }
 
 void ZenFS::ClearFiles() {
-  std::map<std::string, ZoneFile*>::iterator it;
+  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   files_mtx_.lock();
-  for (it = files_.begin(); it != files_.end(); it++) delete it->second;
+  for (it = files_.begin(); it != files_.end(); it++) it->second.reset();
   files_.clear();
   files_mtx_.unlock();
 }
@@ -261,7 +261,7 @@ IOStatus ZenFS::WriteSnapshotLocked(ZenMetaLog* meta_log) {
   s = meta_log->AddRecord(snapshot);
   if (s.ok()) {
     for (auto it = files_.begin(); it != files_.end(); it++) {
-      ZoneFile* zoneFile = it->second;
+      std::shared_ptr<ZoneFile> zoneFile = it->second;
       zoneFile->MetadataSynced();
     }
   }
@@ -357,7 +357,7 @@ IOStatus ZenFS::PersistRecord(std::string record) {
   return s;
 }
 
-IOStatus ZenFS::SyncFileMetadata(ZoneFile* zoneFile) {
+IOStatus ZenFS::SyncFileMetadata(std::shared_ptr<ZoneFile> zoneFile) {
   std::string fileRecord;
   std::string output;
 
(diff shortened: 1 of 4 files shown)


skyzh:
I think #51 is more of a semantics issue of a file system. On ext4 and xfs (and many other Linux file systems), a file can still be accessed when being deleted. It is only until all applications have closed the file that the file system will actually delete the file content.

For ZenFS, we can either implement "delete-after-close" semantics like Linux, or panic when a user delete a being-opened file. This PR can surely avoid the core dump in #51, but it doesn't solve the underlying issue -- it is possible that users may read garbage data or wrong data after they delete a file while still holding reference to a ZoneFile.

metaspace:
> For ZenFS, we can either implement "delete-after-close" semantics like Linux, or panic when a user deletes a being-opened file. This PR can surely avoid the core dump in #51, but it doesn't solve the underlying issue -- it is possible that users may read garbage data or wrong data after they delete a file while still holding reference to a ZoneFile.

@skyzh I think you are right - the holder of a `FSWritableFile` would not observe correct behavior after `ZoneFS::DeleteFile()`. @aravind-wdc  now that we reference count `ZoneFile`, perhaps we can put a delete marker on the `ZoneFile` in `ZoneFS::DeleteFile()` and then perform the deletion in `ZoneFile::~ZoneFile()`?

yhr:
@skyzh : How could garbage/wrong data be read here? The zone file will not deallocate its extents until the ref count = 0 (all readers are done)

skyzh:
* When `zoneFile.reset()` is called, the destructor is executed https://github.com/westerndigitalcorporation/zenfs/pull/76/files#diff-b12ba0b7fd2feee53dba42938e4031b32be21c71dc948ba759844c4c49be0a93R416
* and the extents are then deleted at https://github.com/westerndigitalcorporation/zenfs/blob/d318f03845fc023596e62e949e2fea7f7cac905e/fs/io_zenfs.cc#L221

Perhaps we can simply remove `.reset()` in this PR.

fs/fs_zenfs.cc
```diff
@@ -379,24 +379,24 @@ IOStatus ZenFS::SyncFileMetadata(ZoneFile* zoneFile) {
 }
 
 /* Must hold files_mtx_ */
-ZoneFile* ZenFS::GetFileInternal(std::string fname) {
-  ZoneFile* zoneFile = nullptr;
+std::shared_ptr<ZoneFile> ZenFS::GetFileInternal(std::string fname) {
+  std::shared_ptr<ZoneFile> zoneFile(nullptr);
```
yhr, line 383: Why do we need a separate method for this?

And the whole point of the get is to increase the ref count, right? We need to return zoneFile.get()

yhr (changes requested):

yhr:
@skyzh : The zone file is destroyed when the last user holding a reference to the zone file is deleted. Or did i miss something here?

yhr:
@metaspace : What incorrect behavior?

skyzh:
> @skyzh : Since zone file is a shared pointer, the zone file is destroyed when the last user holding a reference to the zone file is deleted. Or did i miss something here?

Calling .reset on a shared ptr immediately calls the dtor, if I'm understanding correctly.

yhr:
@skyzh : https://en.cppreference.com/w/cpp/memory/shared_ptr

std::shared_ptr is a smart pointer that retains shared ownership of an object through a pointer. Several shared_ptr objects may own the same object. The object is destroyed and its memory deallocated when either of the following happens:

**the last remaining shared_ptr owning the object is destroyed;
the last remaining shared_ptr owning the object is assigned another pointer via operator= or reset().**

metaspace:
> @metaspace : What incorrect behavior?

@yhr I was thinking that executing the logic in `~ZoneFile`, which is what we do on master, would free the extents and cause trouble for readers. But with this PR we will not run that code until the last reader lets go of a `ZoneFile`, so it is not an issue.

skyzh:
Thanks for the explanation, I missed the "last remaining shared_ptr" part, and this really make sense. By the way, it seems unnecessary to `reset` the pointer by ourselves. Upon going out of the function scope, it will be automatically destroyed :)

yhr:
@skyzh : yep, this is what shared pointers is all about :)... and this change makes the the zone file handling a lot more robust - zero chance that a reader will try to access a deallocated zone file

yhr:
Shared pointers are a bit magical, and it's not obvious how they work. I don't like the overloading of the = sign for example, but that's just me (and i'm no c++ guy)

fs/fs_zenfs.cc
aravind-wdc: It was done to avoid race condition in https://github.com/westerndigitalcorporation/zenfs/pull/25

aravind-wdc:
Ohh! Lot of discussion happened. I had to go offline for a couple of hours..
It looks like most of the questions are answered, the object is not deleted until the ref count is reached 0, even if reset is called.

yhr (approved):
Looks good to me. We could drop the explicit .resets(), but I don't mind keeping them as it makes the code easier to understand imho.

yhr:
@metaspace : are you happy with the patch as well?

fs/fs_zenfs.cc
```diff
@@ -245,9 +245,9 @@ void ZenFS::LogFiles() {
 }
 
 void ZenFS::ClearFiles() {
-  std::map<std::string, ZoneFile*>::iterator it;
+  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   files_mtx_.lock();
-  for (it = files_.begin(); it != files_.end(); it++) delete it->second;
+  for (it = files_.begin(); it != files_.end(); it++) it->second.reset();
```
metaspace, line 250: This line is redundant. All objects in the map are destroyed when we call `files_.clear()` on the next line.

fs/fs_zenfs.cc
```diff
@@ -471,13 +470,14 @@ IOStatus ZenFS::NewWritableFile(const std::string& fname,
...
-  zoneFile = new ZoneFile(zbd_, fname, next_file_id_++);
+  std::shared_ptr<ZoneFile> zoneFile(
+      new ZoneFile(zbd_, fname, next_file_id_++));
   zoneFile->SetFileModificationTime(time(0));
 
   /* Persist the creation of the file */
   s = SyncFileMetadata(zoneFile);
   if (!s.ok()) {
-    delete zoneFile;
+    zoneFile.reset();
```
metaspace, line 480: This reset is redundant

metaspace (approved):
I added some comments on redundant lines. No strong feelings about it though. Otherwise it looks good to me.

PR #78, opened 2021-11-04
Add busy flag to `Zone`

metaspace:
This patch set prevents a race condition on `Zone::open_for_write_` by adding a busy
flag to the Zone structure. Writing the busy flag through class methods use
acquire/release ordering, while reading the flag uses relaxed ordering. That is, claiming or releasing a zone is a synchronization operation that involves previous and next owners of the zone.

With this patch, users of a zone must acquire ownership of the zone by setting the busy flag of the zone, before managing the zone.

The patch removes `Zone::open_for_write_` as it becomes redundant.

Supersedes #71 
Supersedes #46

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -277,8 +277,8 @@ IOStatus ZenFS::WriteEndRecord(ZenMetaLog* meta_log) {
 
 /* Assumes the files_mtx_ is held */
 IOStatus ZenFS::RollMetaZoneLocked() {
-  ZenMetaLog* new_meta_log;
-  Zone *new_meta_zone, *old_meta_zone;
+  std::unique_ptr<ZenMetaLog> new_meta_log, old_meta_log;
+  Zone* new_meta_zone;
   IOStatus s;
 
   new_meta_zone = zbd_->AllocateMetaZone();
@@ -289,16 +289,18 @@ IOStatus ZenFS::RollMetaZoneLocked() {
   }
 
   Info(logger_, "Rolling to metazone %d\n", (int)new_meta_zone->GetZoneNr());
-  new_meta_log = new ZenMetaLog(zbd_, new_meta_zone);
+  new_meta_log.reset(new ZenMetaLog(zbd_, new_meta_zone));
 
-  old_meta_zone = meta_log_->GetZone();
-  old_meta_zone->open_for_write_ = false;
+  old_meta_log.swap(this->meta_log_);
+  this->meta_log_.swap(new_meta_log);
+
+  old_meta_log->GetZone()->open_for_write_ = false;
 
   /* Write an end record and finish the meta data zone if there is space left */
-  if (old_meta_zone->GetCapacityLeft()) WriteEndRecord(meta_log_.get());
-  if (old_meta_zone->GetCapacityLeft()) old_meta_zone->Finish();
-
-  meta_log_.reset(new_meta_log);
+  if (old_meta_log->GetZone()->GetCapacityLeft())
+    WriteEndRecord(old_meta_log.get());
+  if (old_meta_log->GetZone()->GetCapacityLeft())
+    old_meta_log->GetZone()->Finish();
 
   std::string super_string;
   superblock_->EncodeTo(&super_string);
@@ -313,7 +315,7 @@ IOStatus ZenFS::RollMetaZoneLocked() {
   s = WriteSnapshotLocked(meta_log_.get());
 
   /* We've rolled successfully, we can reset the old zone now */
-  if (s.ok()) old_meta_zone->Reset();
+  if (s.ok()) old_meta_log->GetZone()->Reset();
 
   return s;
 }
@@ -875,6 +877,7 @@ Status ZenFS::Mount(bool readonly) {
     std::string scratch;
     Slice super_record;
 
+    assert(z->SetBusy());
     log.reset(new ZenMetaLog(zbd_, z));
 
     if (!log->ReadRecord(&super_record, &scratch).ok()) continue;
@@ -1004,8 +1007,13 @@ Status ZenFS::MkFS(std::string aux_fs_path, uint32_t finish_threshold) {
   zbd_->ResetUnusedIOZones();
 
   for (const auto mz : metazones) {
+    assert(mz->SetBusy());
     if (mz->Reset().ok()) {
-      if (!meta_zone) meta_zone = mz;
+      if (!meta_zone) {
+        meta_zone = mz;
+      } else {
+        assert(mz->UnsetBusy());
+      }
     } else {
       Warn(logger_, "Failed to reset meta zone\n");
     }
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -875,7 +875,7 @@ Status ZenFS::Mount(bool readonly) {
     std::string scratch;
     Slice super_record;
 
-    bool ok = z->SetBusy();
+    bool ok = z->Acquire();
     assert(ok);
     _unused(ok);
 
@@ -1008,7 +1008,7 @@ Status ZenFS::MkFS(std::string aux_fs_path, uint32_t finish_threshold) {
   zbd_->ResetUnusedIOZones();
 
   for (const auto mz : metazones) {
-    bool ok = mz->SetBusy();
+    bool ok = mz->Acquire();
     assert(ok);
     _unused(ok);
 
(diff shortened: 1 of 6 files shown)


metaspace:
New version to make the linter happy

floatious:
Perhaps give some credit to
https://github.com/westerndigitalcorporation/zenfs/pull/46
by adding a:
Reported-by: Kuankuan Guo guokuankuan@bytedance.com

If it was also reported by Percona, you can add another Reported-by tag
with whoever that found it, so that you have two Reported-by tags.

And maybe overkill, but since Hans suggested that you should use a busy flag,
you could also add a:
Suggested-by: Hans Holmberg <hans.holmberg@wdc.com>

fs/io_zenfs.cc
```diff
@@ -361,7 +361,7 @@ IOStatus ZoneFile::Append(void* data, int data_size, int valid_size) {
   if (!active_zone_) {
     Zone* zone = zbd_->AllocateZone(lifetime_);
     if (!zone) {
-      return IOStatus::NoSpace("Zone allocation failure\n");
+      return IOStatus::NoSpace("Zone allocation failure, no active zone\n");
```
aravind-wdc, line 364: We could also hit the "No Space" when all the zones are actually full and active zones are less than max-active-zones.  

Would it be better to dump more information such as active-zone-count, full-zone-count, total-zone-count and let the user make sense out of it ?

fs/zbd_zenfs.cc
```diff
@@ -544,20 +579,33 @@ Zone *ZonedBlockDevice::AllocateZone(Env::WriteLifeTimeHint file_lifetime) {
 
     if (active_io_zones_.load() < max_nr_active_io_zones_) {
       for (const auto z : io_zones) {
-        if ((!z->open_for_write_) && z->IsEmpty()) {
-          z->lifetime_ = file_lifetime;
-          allocated_zone = z;
-          active_io_zones_++;
-          new_zone = 1;
-          break;
+        if (z->SetBusy()) {
```
aravind-wdc, line 580: IIRC, SetBusy() would return true if it goes and set busy_ as true, else it returns false.
In case it does set it to true and if the zone is not empty(say it is a closed zone), then we will be setting this zone as busy and unset it in the else part below . 

Wouldn't it be better if we do a if ( ! IsBusy()) here and then actually do SetBusy() when the zone is allocated.

fs/fs_zenfs.cc
```diff
@@ -277,8 +277,8 @@ IOStatus ZenFS::WriteEndRecord(ZenMetaLog* meta_log) {
 
 /* Assumes the files_mtx_ is held */
 IOStatus ZenFS::RollMetaZoneLocked() {
-  ZenMetaLog* new_meta_log;
-  Zone *new_meta_zone, *old_meta_zone;
+  std::unique_ptr<ZenMetaLog> new_meta_log, old_meta_log;
```
aravind-wdc, line 280: Do you think it would be better to split the metalog related changes to another patch ?

fs/io_zenfs.cc
metaspace: This one just says zone allocation failure, nothing about how much space is left. I think caller may add the out of space message. I added this text to distinguish the error from a very similar one in line 384.

Regarding the zone stats, it is available if running with debug. In production mode I think it should be the responsibility of the caller to trigger printing of the stats, if desired.

fs/zbd_zenfs.cc
metaspace: `Zone::IsBusy()` has relaxed memory ordering, so none of the fields of `Zone` is guaranteed to be observed to the value they were set to, before the last owner of the zone called `Zone::UnsetBusy()`. That is to make `Zone::IsBusy()` faster.

We could change `Zone::IsBusy()` to stronger memory ordering, but we would still have a race on `Zone::lifetime_`.

Another approach could be to make a lockless implementation where the `Zone` pointer is atomically updated in-place in the io_zones list with a compare and swap operation. You would have to do this update every time you wanted to change a field of a `Zone`, and you would pass zones around as indexes to the list instead of as pointers. You would still need acquire/release ordering for the compare and swap operation though.

Could be interesting to benchmark the two approaches against each other.

fs/fs_zenfs.cc
metaspace: I can do that, I have no opinion on that. But the system would be broken after application of the first patch, before the second one.

fs/io_zenfs.cc
yhr: I think that "No active zone" adds no value to the end user. "Out of space: zone allocation failure" would be helpful.

fs/zbd_zenfs.cc
yhr: I think we should prioritize that the code should be a) robust and b) easy to understand. The first six rules of optimizations are: don't do it.

As long as Unset/SetBusy is safe to be accessed from different thtreads, i'm good.

fs/io_zenfs.cc
metaspace: I see that I missed something. Out of space is already added, it is an `IOStatus::NoSpace`. No reason to put it in the message.

I made the change because identical errors were raised in multiple locations. For debugging, it was impossible to know where the error originated. We need either unique error messages, or we need to annotate the error with file name and line number.

fs/zbd_zenfs.cc
```diff
@@ -513,22 +529,41 @@ Zone *ZonedBlockDevice::AllocateZone(Env::WriteLifeTimeHint file_lifetime) {
...
+      if ((z->used_capacity_ > 0) && !z->IsFull()) {
+        unsigned int diff = GetLifeTimeDiff(z->lifetime_, file_lifetime);
+        if (diff <= best_diff) {
+          if (allocated_zone != nullptr) {
+            assert(allocated_zone->UnsetBusy());
+          }
+          allocated_zone = z;
+          best_diff = diff;
+        } else {
+          assert(z->UnsetBusy());
```
yhr, line 554: Danger danger! everything inside assert() will be removed when built DEBUG_LEVEL==0 , not what we want, right?

fs/io_zenfs.cc
yhr: I think it is motivated to add "out of space" to the message as I recall people reporting "Zone allocation error" to me via email and not understanding that they actually ran out of space..

If we have different allocation errors, we should return what error we hit when returning from AllocateZone

metaspace: Printing a `IOStatus::NoSpace` error now gives something like `NoSpace: <message>`. Adding 'out of space' would print `Error ... NoSpace: Zone allocation error, no space`. Do we want that, it is sort of redundant?

> If we have different allocation errors, we should return what error we hit when returning from AllocateZone
That is a good idea, I will add that.

But we still call `AllocateZone` in multiple locations. For debugging, it is desirable to know what call site is propagating the error.

fs/zbd_zenfs.cc
metaspace: > I think we should prioritize that the code should be a) robust and b) easy to understand. 

> As long as Unset/SetBusy is safe to be accessed from different thtreads, i'm good.

I Agree. The current name of the methods does not really convey that you grab a mutex when you call them. How about renaming to `Acquire()` and `Release()` ?

fs/zbd_zenfs.cc
metaspace: Good catch! Will change that. This also reveals that I only ran the test suite with debug mode enabled :confounded:

fs/zbd_zenfs.cc
yhr: Sounds good

fs/zbd_zenfs.cc
MaisenbacherD: This PR is addressing the problem: https://github.com/metaspace/zenfs/pull/1

metaspace:
@yhr Pushed new commits to address the issues mentioned in the comments. I think we can defer adding metaspace/zenfs#1 until we have more bandwidth.

yhr (approved):
Looks good to me, i'll squash the patches into a single commit and push it manually. @MaisenbacherD   work on top of this can go as a separate pull request to master

metaspace:
Not sure I use the draft/ready feature the right way :man_shrugging:

yhr:
If it's ready for review then it should not be in "draft", so i assumed that this was really in non-draft mode :)

I'm now squashing and fixing up one remaining thing (the pesky this->'s ). I'll make another build & smoke test before pushing

yhr:
Squased & Pushed in 5dd349c0235782785bc472ff7344d9eed924be77

PR #81, opened 2021-11-15
CI smoke test

metaspace:
Add a github action to run the ZenFS smoke tests.

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -1,0 +1,31 @@
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+name: Smoke Test
+jobs:
+  smoke-test:
+    runs-on: self-hosted
+    steps:
+      - name: Clean
+        run: rm -rf ${GITHUB_WORKSPACE}/*
+      - name: Checkout build scripts
+        run: git clone ~/harness.git
+      - name: Checkout zbdbench
+        run: git clone ~/zbdbench.git harness/rocksdb-context/zbdbench
+      - name: Checkout rocksdb
+        uses: actions/checkout@v2
+        with:
+          repository: facebook/rocksdb
+          ref: v6.25.3
+          path: harness/rocksdb-context/rocksdb
+      - name: Checkout zenfs
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/smoke-test.yaml
```diff
@@ -0,0 +1,31 @@
...
+  smoke-test:
+    runs-on: self-hosted
+    steps:
+      - name: Clean
+        run: rm -rf ${GITHUB_WORKSPACE}/*
+      - name: Checkout build scripts
+        run: git clone ~/harness.git
+      - name: Checkout zbdbench
+        run: git clone ~/zbdbench.git harness/rocksdb-context/zbdbench
+      - name: Checkout rocksdb
```
yhr, line 19: I don't know how the harness works, but can't we have all of the smoke-test specifics in that project? (rocksdb version, how we run the tests)  - now it looks like we have to change both the harness AND this yaml if we want to change how the smoke test is run.

yhr (changes requested):
See my comment, looks good apart from that. Also please prefix your commit messages with CI: for ci-stuff, and tests: for changes to the testing scripts

skyzh:
I think this is a little bit dangerous. Anyone sending PRs could run anything on your physical machine, including accessing your internal infrastructure, gaining access to the underlying hardware, etc. See https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories for more information:

> Untrusted workflows running on your self-hosted runner pose significant security risks for your machine and network environment, especially if your machine persists its environment between jobs. Some of the risks include:
> 
> * Malicious programs running on the machine.
> * Escaping the machine's runner sandbox.
> * Exposing access to the machine's network environment.
> * Persisting unwanted or dangerous data on the machine.


I think this is better to be part of WD's internal daily CI/CD procedure, instead of something public on GitHub.

.github/workflows/smoke-test.yaml
metaspace: Yes, those things live in the test harness. The harness runs the smoke test with default settings. Parameters are not intended to be modified beyond default for this CI check, so there is really nothing to change.

The harness is not public at the moment, but I will look into releasing it, if it makes sense.

PR #85, opened 2021-11-17
tests: print and store time spent for tests and test sets

yhr:
It's useful to know how long a test runs, both for scheduling
testing and to compare different runs to look for regressions
and improvements in execution time.

So, print the time it takes to run each test and the total
time to run a whole test set in run.sh. Also store the time in the results
directory for comparing different runs.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -4,6 +4,15 @@ set -e
 SCRIPT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)
 cd $SCRIPT_DIR
 
+function print_duration() {
+  SECS=$1
+  HRS=$(($SECS / 3600))
+  SECS=$(($SECS % 3600))
+  MINS=$(($SECS / 60))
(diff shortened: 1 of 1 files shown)


MaisenbacherD:
Looks good to me.

tests/run.sh
```diff
@@ -18,20 +27,24 @@ RESULT_DIR="results/$NAME/$TEST_DIR"
 mkdir -p $RESULT_DIR
 rm -rf $RESULT_DIR/*
 
+SECONDS=0
```
metaspace, line 30: Could you make a comment (maybe with a reference to the manual) about how this variable works?

metaspace (changes requested):

tests/run.sh
yhr: yeah, it's a bit obscure

yhr: fixed

metaspace (approved):

yhr:
Rebased

PR #87, opened 2021-11-17
util: zenfs mkfs does not create the aux-dir immediately

aravind-wdc:
Presently aux-dir is created at mount time, but it should be
created at mkfs time. During mkfs, create aux-dir if it is not present.
Check if it is empty and accessible by the user, else error out.

### DIFF
--- a/util/zenfs.cc
+++ b/util/zenfs.cc
@@ -12,6 +12,7 @@
 #include <sys/types.h>
 
 #include <cstdio>
+#include <cstring>
 #include <fstream>
 #include <iostream>
 #include <memory>
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
   return s;
 }
 
+int is_dir(const char *path) {
+  struct stat st;
+  if (stat(path, &st) != 0) {
+    fprintf(stderr, "Failed to stat %s\n", path);
+    return 1;
+  }
+  return S_ISDIR(st.st_mode);
+}
+
+// Create or check pre-existing aux directory and fail if it is
+// inaccessible by current user and if it has previous data
+int create_aux_dir(const char *path) {
+  struct dirent *dent;
+  size_t nfiles = 0;
+  int ret = 0;
+
+  errno = 0;
+  ret = mkdir(path, 0750);
+  if (ret < 0 && EEXIST != errno) {
+    fprintf(stderr, "Failed to create aux directory %s: %s\n", path,
+            strerror(errno));
+    return 1;
+  }
+  // The aux_path is now available, check if it is a directory infact
+  // and is empty and the user has access permission
+
+  if (!is_dir(path)) {
+    fprintf(stderr, "Aux path %s is not a directory\n", path);
+    return 1;
+  }
+
+  errno = 0;
+
+  auto closedirDeleter = [](DIR *d) {
+    if (d != nullptr) closedir(d);
+  };
+  std::unique_ptr<DIR, decltype(closedirDeleter)> aux_dir{
+      opendir(path), std::move(closedirDeleter)};
+  if (errno) {
+    fprintf(stderr, "Failed to open aux directory %s: %s\n", path,
+            strerror(errno));
+    return 1;
+  }
+
+  // Consider the directory as non-empty if any files/dir other
+  // than . and .. are found.
+  while ((dent = readdir(aux_dir.get())) != NULL && nfiles <= 2) ++nfiles;
+  if (nfiles > 2) {
+    fprintf(stderr, "Aux directory %s is not empty.\n", path);
+    return 1;
+  }
+
+  if (access(path, R_OK | W_OK | X_OK) < 0) {
+    fprintf(stderr,
+            "User does not have access permissions on "
+            "aux directory %s\n",
+            path);
+    return 1;
+  }
+
+  return 0;
+}
+
 int zenfs_tool_mkfs() {
   Status s;
-  DIR *aux_dir;
 
   if (FLAGS_aux_path.empty()) {
     fprintf(stderr, "You need to specify --aux_path\n");
     return 1;
   }
 
-  aux_dir = opendir(FLAGS_aux_path.c_str());
-  if (ENOENT != errno) {
-    fprintf(stderr, "Error: aux path exists\n");
-    closedir(aux_dir);
-    return 1;
-  }
+  if (create_aux_dir(FLAGS_aux_path.c_str())) return 1;
 
   std::unique_ptr<ZonedBlockDevice> zbd = zbd_open(false, true);
   if (!zbd) return 1;


aravind-wdc:
```
$ ./zenfs mkfs --zbd=nvme2n2 --aux_path=/var/lib/zns11 --force
ZenFS: Error creating directory /var/lib/zns11: Permission denied

Run db_bench
$ sudo ./zenfs mkfs --zbd=nvme2n2 --aux_path=/var/lib/zns11 --force
ZenFs: Directory /var/lib/zns11 not empty.

Create a dir /var/lib/zns5 as root.
$ sudo -u wdc ./zenfs mkfs --zbd=nvme2n2 --aux_path=/var/lib/zns5 --force
ZenFS: User does not have access permisisons on /var/lib/zns5

$ sudo ./zenfs mkfs --zbd=nvme2n2 --aux_path=/var/lib/zns5 --force
ZenFS file system created. Free space: 3968745 MB

```

util/zenfs.cc
```diff
@@ -69,21 +70,56 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+            FLAGS_aux_path.c_str(), strerror(errno));
+    return 1;
+  }
+  // The aux_path is now available, check if it is not empty and
+  // the user has access permission
+
+  errno = 0;
+  aux_dir = opendir(FLAGS_aux_path.c_str());
+  if (errno) {
+    fprintf(stderr, "ZenFS: Error opening directory %s: %s\n",
```
yhr, line 94: You can drop the ZenFS prefix(es), in the error messages, it is not needed. In stead, add that the aux directory could not be opened/is not empty/ does not have the right permissions

util/zenfs.cc
```diff
@@ -69,21 +70,56 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+            FLAGS_aux_path.c_str(), strerror(errno));
+    return 1;
+  }
+
+  while ((dent = readdir(aux_dir)) != NULL && nfiles <= 2) nfiles++;
+  if (nfiles > 2) {
```
yhr, line 100: Please add a comment on this condition, why > 2?

yhr (changes requested):
Looks good to me except for a couple of nitpick. I'll ask yura to review as well to make sure we solve the issue he reported

util/zenfs.cc
```diff
@@ -69,21 +70,56 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+  if (ret < 0 && EEXIST != errno) {
+    fprintf(stderr, "ZenFS: Error creating directory %s: %s\n",
+            FLAGS_aux_path.c_str(), strerror(errno));
+    return 1;
+  }
+  // The aux_path is now available, check if it is not empty and
+  // the user has access permission
+
+  errno = 0;
+  aux_dir = opendir(FLAGS_aux_path.c_str());
```
metaspace, line 92: I think that we should pass in the path as an argument, rather than relying on global state. Makes it easier to test the method.

util/zenfs.cc
```diff
@@ -69,21 +70,56 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+// Create or check pre-existing aux directory and fail if it is
+// inaccessible by current user and if it has previous data
+int create_aux_path() {
+  struct dirent *dent;
+  DIR *aux_dir;
+  unsigned int nfiles = 0;
+  int ret = 0;
+
+  errno = 0;
+  ret = mkdir(FLAGS_aux_path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
```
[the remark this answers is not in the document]
percona-ysorokin: I would suggest the default permissions for this directory to be `750`.
Owner (`mysql` user) is allowed to read/write
Group(for instance backup users who belong to `mysql` group) is allowed to read.
The directory is inaccessible to others.

util/zenfs.cc
```diff
@@ -8,6 +8,7 @@
 #include <gflags/gflags.h>
 #include <rocksdb/file_system.h>
 #include <rocksdb/plugin/zenfs/fs/fs_zenfs.h>
+#include <string.h>
```
percona-ysorokin, line 11: Use c++ headers if possible
```suggestion
#include <cstring>
```

util/zenfs.cc
percona-ysorokin, line 92: Put the result of the `opendir()` into a RAII object to avoid memory leaks
```suggestion
  auto closedirDeleter = [](DIR* d) {
    if (d != nullptr) closedir(d);
  };
  std::unique_ptr<DIR, decltype(closedirDeleter)> aux_dir{
      FLAGS_aux_path.c_str(), std::move(closedirDeleter)};
```

util/zenfs.cc
percona-ysorokin: Another corner case is reading the content of the `/` (root directory) - in this case we won't have `.` and `..` in the output.
Although I must admit that specifying `aux_dir=/` looks unrealistic to me (may only be meaningful under something like `chroot`).

util/zenfs.cc
```diff
@@ -69,21 +70,56 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+
+  errno = 0;
+  aux_dir = opendir(FLAGS_aux_path.c_str());
+  if (errno) {
+    fprintf(stderr, "ZenFS: Error opening directory %s: %s\n",
+            FLAGS_aux_path.c_str(), strerror(errno));
+    return 1;
+  }
+
+  while ((dent = readdir(aux_dir)) != NULL && nfiles <= 2) nfiles++;
```
percona-ysorokin, line 99: Modern c++ style
```suggestion
  while ((dent = readdir(aux_dir)) != nullptr && nfiles <= 2) ++nfiles;
```

util/zenfs.cc
```diff
@@ -69,21 +70,56 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
   return s;
 }
 
+// Create or check pre-existing aux directory and fail if it is
+// inaccessible by current user and if it has previous data
+int create_aux_path() {
+  struct dirent *dent;
+  DIR *aux_dir;
+  unsigned int nfiles = 0;
```
percona-ysorokin, line 78: almost always `std::size_t` for counters
```suggestion
  std::size_t nfiles = 0;
```

util/zenfs.cc
```diff
@@ -69,21 +70,56 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+  int ret = 0;
+
+  errno = 0;
+  ret = mkdir(FLAGS_aux_path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
+  if (ret < 0 && EEXIST != errno) {
+    fprintf(stderr, "ZenFS: Error creating directory %s: %s\n",
+            FLAGS_aux_path.c_str(), strerror(errno));
+    return 1;
+  }
+  // The aux_path is now available, check if it is not empty and
```
percona-ysorokin, line 88: One more corner case here.
What happens if `aux_path` exists but it is a file rather than a directory.
Another question is a `symlink` pointing to either directory (`OK`) or file (`not OK`).

aravind-wdc: Done.

util/zenfs.cc
aravind-wdc: Ok

util/zenfs.cc
aravind-wdc: Ok.

util/zenfs.cc
aravind-wdc: Given zenfs utility is a standalone and small application, i don't think there is threat of any kind of huge memory leaks.
So I feel that this is a bit over-engineering.
Probably we can make a separate commit introducing such changes in all places.

util/zenfs.cc
aravind-wdc: Ok.

util/zenfs.cc
aravind-wdc: Done.

util/zenfs.cc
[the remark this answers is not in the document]
aravind-wdc: I thought user group should have the perms. But 0750 also make sense.
Changed to 0750 now.

util/zenfs.cc
aravind-wdc: Added the comment.
I can see  . and .. when i do a "ls -a /" , and anyway mkfs will fail when run on /, unless it has less than 2 files, in which case it is already in a bad shape.

util/zenfs.cc
aravind-wdc: Ok, done.

util/zenfs.cc
metaspace: I don't think code should be allowed to leak because it is used in the context of a small program. You never know how a piece of code will be used in the future. It's just good practice to make all code well behaved. 

I think it would be proper to call `closedir()` on the `DIR` before returning from the function. If you don't want to use RAII, just do it everywhere you return. Using a `unique_ptr` is cleaner in my opinion.

aravind-wdc: I had totally missed that I had not done closedir() and was talking about any hypothetical leaks, of course nothing should be allowed to leak when it is apparent. Updated to do closedir() now.

util/zenfs.cc
```diff
@@ -69,22 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+  if (access(FLAGS_aux_path.c_str(), R_OK | W_OK | X_OK) < 0) {
+    fprintf(stderr,
+            "User does not have access permissions on "
+            "aux directory %s\n",
+            FLAGS_aux_path.c_str());
+    closedir(aux_dir);
+    return 1;
+  }
+
+  return 0;
```
metaspace, line 132: I think we need `closedir()` here as well, right?

metaspace (changes requested):

util/zenfs.cc
aravind-wdc: You are right, thanks, added it.

metaspace (approved):

util/zenfs.cc
percona-ysorokin: OK, then

util/zenfs.cc
percona-ysorokin: @aravind-wdc This is exactly why I suggested using RAII, instead of 3 manual calls to `closedir()` before each `return` (one of which was missing), you declare everything in one place and delegate the responsibility of releasing allocated resources to the compiler.

metaspace: @percona-ysorokin completely agree, this is a perfect demonstration of why RAII is awesome :)

util/zenfs.cc
aravind-wdc: Updated to use the RAII object to track the DIR* aux_dir in stead of regular old pointer.

util/zenfs.cc
```diff
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+
+// Create or check pre-existing aux directory and fail if it is
+// inaccessible by current user and if it has previous data
+int create_aux_dir(const char *path) {
+  struct dirent *dent;
+  size_t nfiles = 0;
+  int ret = 0;
+
+  errno = 0;
+  ret = mkdir(FLAGS_aux_path.c_str(), 0750);
```
metaspace, line 90:
```suggestion
  ret = mkdir(path, 0750);
```

util/zenfs.cc
```diff
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+    return 1;
+  }
+
+  errno = 0;
+
+  auto closedirDeleter = [](DIR *d) {
+    if (d != nullptr) closedir(d);
+  };
+  std::unique_ptr<DIR, decltype(closedirDeleter)> aux_dir{
+      opendir(FLAGS_aux_path.c_str()), std::move(closedirDeleter)};
```
metaspace, line 110:
```suggestion
      opendir(path), std::move(closedirDeleter)};
```

util/zenfs.cc
```diff
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+  if (errno) {
+    fprintf(stderr, "Failed to open aux directory %s: %s\n",
+            FLAGS_aux_path.c_str(), strerror(errno));
```
metaspace, line 113:
```suggestion
            path, strerror(errno));
```

util/zenfs.cc
```diff
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+int create_aux_dir(const char *path) {
+  struct dirent *dent;
+  size_t nfiles = 0;
+  int ret = 0;
+
+  errno = 0;
+  ret = mkdir(FLAGS_aux_path.c_str(), 0750);
+  if (ret < 0 && EEXIST != errno) {
+    fprintf(stderr, "Failed to create aux directory %s: %s\n",
+            FLAGS_aux_path.c_str(), strerror(errno));
```
metaspace, line 93:
```suggestion
            path, strerror(errno));
```

util/zenfs.cc
```diff
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+    fprintf(stderr, "Failed to open aux directory %s: %s\n",
+            FLAGS_aux_path.c_str(), strerror(errno));
+    return 1;
+  }
+
+  // Consider the directory as non-empty if any files/dir other
+  // than . and .. are found.
+  while ((dent = readdir(aux_dir.get())) != NULL && nfiles <= 2) ++nfiles;
+  if (nfiles > 2) {
+    fprintf(stderr, "Aux directory %s is not empty.\n", FLAGS_aux_path.c_str());
```
metaspace, line 121:
```suggestion
    fprintf(stderr, "Aux directory %s is not empty.\n", path);
```

util/zenfs.cc
```diff
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+    return 1;
+  }
+
+  if (access(FLAGS_aux_path.c_str(), R_OK | W_OK | X_OK) < 0) {
```
metaspace, line 125:
```suggestion
  if (access(path, R_OK | W_OK | X_OK) < 0) {
```

util/zenfs.cc
```diff
@@ -69,21 +70,78 @@ Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
...
+    fprintf(stderr,
+            "User does not have access permissions on "
+            "aux directory %s\n",
+            FLAGS_aux_path.c_str());
```
metaspace, line 129:
```suggestion
            path);
```

metaspace (changes requested):

util/zenfs.cc
aravind-wdc: Looks like I am on a roll, on this PR :)
Updated the path.

yhr:
@percona-ysorokin : If you're happy with the latest update, lets merge this :)

metaspace (approved):

percona-ysorokin:
@yhr Let me build everything and test the scenario I initially reported (the one with the `aux_dir` inside `/var/lib`). I will let you know as soon as I have the results. Everything else looks good to me.

percona-ysorokin:
@yhr All my manual testing was successful. Please, proceed with merging.

yhr:
thanks for reviewing & testing @percona-ysorokin

yhr (approved):

PR #88, opened 2021-11-19
Update error messages raised when open device fail

metaspace:
Make error messages unique for open device errors.


yhr:
Looks good to me, but please update the commit message to say "fs: make error messages unique for open device errors."

yhr (approved):

PR #89, opened 2021-11-19
Allow extra compiler/linker flags when building zenfs utility

metaspace:
When libzbd and librocksdb is not living in the same location, it is nice to be able to pass extra flags to the build script.

### DIFF
--- a/util/Makefile
+++ b/util/Makefile
@@ -5,13 +5,16 @@ TARGET = zenfs
 CC ?= gcc
 CXX ?= g++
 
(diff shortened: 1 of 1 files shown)


yhr:
please add "util:" as a prefix for utility fixes, otherwise this looks good to me

metaspace:
This PR is originally to address an issue that occur when librocksdb and libzbd is installed to different locations. I have made a PR to address the issue in rocksdb: facebook/rocksdb#9198. However, I think this change still has merit, if somebody wants to compile with other flags for debug or something.

PR #91, opened 2021-11-23
fs: Refactor `ZoneFile` for code reuse

metaspace:
Add a method to close the active zone.

### DIFF
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -229,15 +229,22 @@ ZoneFile::~ZoneFile() {
 IOStatus ZoneFile::CloseWR() {
   IOStatus s = IOStatus::OK();
 
-  if (active_zone_) {
-    s = active_zone_->CloseWR();
-    ReleaseActiveZone();
-    if (s.ok()) {
-      zbd_->NotifyIOZoneClosed();
-    }
-  }
+  s = CloseActiveZone();
   open_for_wr_ = false;
 
+  return s;
+}
+
+IOStatus ZoneFile::CloseActiveZone() {
+  IOStatus s = IOStatus::OK();
+  if (active_zone_) {
+    s = active_zone_->CloseWR();
+    if (!s.ok()) {
+      return s;
+    }
+    ReleaseActiveZone();
+    zbd_->NotifyIOZoneClosed();
+  }
   return s;
 }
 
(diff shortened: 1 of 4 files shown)


metaspace:
Pushed a series of commits to clean up locking around open/active zone management in `ZonedBlockDevice`.

metaspace:
Pushed for new commit message wording.

fs/zbd_zenfs.cc
```diff
@@ -342,12 +340,10 @@ IOStatus ZonedBlockDevice::Open(bool readonly, bool exclusive) {
   return IOStatus::OK();
 }
 
-void ZonedBlockDevice::NotifyIOZoneFull() {
-  active_io_zones_--;
-  zone_resources_.notify_one();
-}
+void ZonedBlockDevice::NotifyIOZoneFull() { active_io_zones_--; }
```
yhr, line 343: Why did you remove the notify here?

metaspace: It is as it says in the commit text. We don't use it.

yhr:
The commit message was not very verbose.

I'd like to keep the notify_one() for future use, the zoned block device
should be notified upon zone state transitions - so this patch can just be
dropped.

metaspace:
Ok I see. I will drop it then.

metaspace:
> I'd like to keep the notify_one() for future use, the zoned block device
> should be notified upon zone state transitions - so this patch can just be
> dropped.

I see, I will drop that patch then.

metaspace:
@yhr Addressed your comments and rebased on master.

fs/io_zenfs.cc
```diff
@@ -229,15 +229,22 @@ ZoneFile::~ZoneFile() {
...
+}
+
+IOStatus ZoneFile::CloseActiveZone() {
+  IOStatus s = IOStatus::OK();
   if (active_zone_) {
-    s = active_zone_->CloseWR();
-    ReleaseActiveZone();
-    if (s.ok()) {
-      zbd_->NotifyIOZoneClosed();
+    active_zone_->CloseWR();
```
yhr, line 241: Doh! You're not checking the return value

yhr (changes requested):

fs/io_zenfs.cc
metaspace: oops

yhr (approved):

PR #92, opened 2021-11-23
fs: ensure extents are pushed

aravind-wdc:
Add Fsync() in ZonedWritableFile destructor so that the extent
information is persisted. In some scenarios it can lead to
files without extent information but with a valid size.


yhr:
An Fsync can fail, so it's not good to handle this in the destructor - and this patch only solves half of the problem anyway (if the whole system crashes, the destructor wont be called). I'll create a separate issue for making sure that the file size matches the extent list on mount.

yhr:
https://github.com/westerndigitalcorporation/zenfs/issues/95 created, closing this

PR #93, opened 2021-11-23
fs: ensure ReopenWritableFile() matches rocksdb definition

aravind-wdc:
RocksDB expects the api ReopenWritableFile() to return an existing
file and if the file is not present, create the file and return it.
Ensure ZenFS api adheres to it.

Signed-off-by: Aravind Ramesh<Aravind.Ramesh@wdc.com>

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -460,35 +460,10 @@ IOStatus ZenFS::NewWritableFile(const std::string& fname,
                                 const FileOptions& file_opts,
                                 std::unique_ptr<FSWritableFile>* result,
                                 IODebugContext* /*dbg*/) {
-  IOStatus s;
-
   Debug(logger_, "New writable file: %s direct: %d\n", fname.c_str(),
         file_opts.use_direct_writes);
 
-  if (GetFile(fname) != nullptr) {
-    s = DeleteFile(fname);
-    if (!s.ok()) return s;
-  }
-
-  std::shared_ptr<ZoneFile> zoneFile(
-      new ZoneFile(zbd_, fname, next_file_id_++));
-  zoneFile->SetFileModificationTime(time(0));
-
-  /* Persist the creation of the file */
-  s = SyncFileMetadata(zoneFile);
-  if (!s.ok()) {
-    zoneFile.reset();
-    return s;
-  }
-
-  files_mtx_.lock();
-  files_.insert(std::make_pair(fname.c_str(), zoneFile));
-  files_mtx_.unlock();
-
-  result->reset(new ZonedWritableFile(zbd_, !file_opts.use_direct_writes,
-                                      zoneFile, &metadata_writer_));
-
-  return s;
+  return OpenWritableFile(fname, file_opts, result, nullptr, false);
 }
 
 IOStatus ZenFS::ReuseWritableFile(const std::string& fname,
@@ -516,16 +491,16 @@ IOStatus ZenFS::FileExists(const std::string& fname, const IOOptions& options,
   }
 }
 
+/* If the file does not exist, create a new one,
+ * else return the existing file
+ */
 IOStatus ZenFS::ReopenWritableFile(const std::string& fname,
-                                   const FileOptions& options,
+                                   const FileOptions& file_opts,
                                    std::unique_ptr<FSWritableFile>* result,
                                    IODebugContext* dbg) {
   Debug(logger_, "Reopen writable file: %s \n", fname.c_str());
 
-  if (GetFile(fname) != nullptr)
-    return NewWritableFile(fname, options, result, dbg);
-
-  return target()->NewWritableFile(fname, options, result, dbg);
+  return OpenWritableFile(fname, file_opts, result, dbg, true);
 }
 
 IOStatus ZenFS::GetChildren(const std::string& dir, const IOOptions& options,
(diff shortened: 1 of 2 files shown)


metaspace:
I am working a very similar issue for `ZenFS::DeleteFile()`. Hoping to post a patch later today.

metaspace:
Here it is: #96

fs/fs_zenfs.cc
```diff
@@ -516,16 +516,23 @@ IOStatus ZenFS::FileExists(const std::string& fname, const IOOptions& options,
...
   Debug(logger_, "Reopen writable file: %s \n", fname.c_str());
 
-  if (GetFile(fname) != nullptr)
+  if (zoneFile == nullptr)
     return NewWritableFile(fname, options, result, dbg);
-
-  return target()->NewWritableFile(fname, options, result, dbg);
+  else
+    result->reset(new ZonedWritableFile(zbd_, !options.use_direct_writes,
+                                        zoneFile, &metadata_writer_));
```
metaspace, line 534: We have some overlap of code in `NewWritableFile()`, `ReopenWritableFile()`. Could we not make both delegate to a protected `OpenFile()`? I think that is what the posix layer is doing, we could model after that.

metaspace (changes requested):

fs/fs_zenfs.cc
aravind-wdc: Thanks, good suggestion. Done.

aravind-wdc:
ReuseWritableFile() is also not as per rocksdb definition, fixed it.

yhr:
@aravind-wdc :What is wrong with the current implementation of ReuseWritableFile? Your patch will make returned writablefiles append at the old file's end.

Reading the posix implementation, ReuseWritableFile() re-opens the current file and **starts overwriting the file at offset 0**.
We can't do overwrites in zenfs, so the current implementation in zenfs deletes the old file and creates a new file with the new name. That's ok. 

ReuseWriteableFile seems to be a bit of a hack to avoid new allocations in regular file systems for performance reasons.

aravind-wdc:
@yhr In the rocksdb file_system.h, for ReuseWritableFile(), it says, 
"Reuse an existing file by renaming it and opening it as writable."
https://github.com/facebook/rocksdb/blob/main/include/rocksdb/file_system.h#L367

ReuseWritableFile() has a new file name and an old-filename as the arguments, fname and old_fname respectively.
As per my understanding, the caller expects it to rename the existing file "old_fname" to "fname" and simply return the newly renamed file "fname".

In current ZenFS implementation of ReuseWritableFile() we check for the new file "fname" instead of the existing file "old_fname", since it is new file name, it will not be found, we return NotFound always. 
If it is found we call NewWritableFile(), which basically deletes it and creates a new file. Since we are checking for a non-existing file(fname), so NewWritableFile() is never called.

This patch, checks if old_file is present, if not, it returns that it is not found.
If found, it renames it to old_fname, and reopens the file and returns it (not delete the file)

Please let me know if I am mistaken, I will recall this patch

aravind-wdc:
If it's ok to delete it, then we need to change the GetFile(fname) to GetFile(old_fname) and retain the same behavior of deleting the file.

metaspace:
> If it's ok to delete it, then we need to change the GetFile(fname) to GetFile(old_fname) and retain the same behavior of deleting the file.

@aravind-wdc There is a default implementation in `FileSystem` that we inherit if we don't implement our own:
```c++
  auto closedirDeleter = [](DIR* d) {
    if (d != nullptr) closedir(d);
  };
  std::unique_ptr<DIR, decltype(closedirDeleter)> aux_dir{
      FLAGS_aux_path.c_str(), std::move(closedirDeleter)};
```

metaspace:
> @aravind-wdc :What is wrong with the current implementation of ReuseWritableFile?

@yhr The method is supposed to delete `old_fname`. Our implementation does not do that. Its bugged:
```c++
if (GetFile(fname) == nullptr)
  return IOStatus::NotFound("Old file does not exist");

return NewWritableFile(fname, file_opts, result, dbg);
```

metaspace:
> This patch, checks if old_file is present, if not, it returns that it is not found.
> If found, it renames it to old_fname, and reopens the file and returns it (not delete the file)

@aravind-wdc That looks good to me

aravind-wdc:
> @aravind-wdc What is the actual required semantics? Should the file be truncated after return from `ReuseWritableFile` or should writes continue from the old write pointer?

The actual semantics are not very clear. In the rocksdb/include/file_system.h , it says that the file has to be renamed and returned. But not whether it should be truncated.
I don't think the posix implementation truncates it, i need to check again.
I thought it should continue writing from old WP, until Hans mentioned that it has to be truncated.

aravind-wdc:
> ```c++
> IOStatus s = RenameFile(old_fname, fname, opts.io_options, dbg);
> if (!s.ok()) {
>   return s;
> }
> return NewWritableFile(fname, opts, result, dbg);
> ```

This is what I have done, along with a error log saying "rename failed"

yhr:
> > This patch, checks if old_file is present, if not, it returns that it is not found.
> > If found, it renames it to old_fname, and reopens the file and returns it (not delete the file)
> 
> @aravind-wdc What is the actual required semantics? Should the file be truncated after return from `ReuseWritableFile` or should writes continue from the old write pointer?
> 
> > Reading the posix implementation, ReuseWritableFile() re-opens the current file and starts overwriting the file at offset 0.
> 
> @yhr where do you see this? It is not in the default implementation in `FileSystem::ReuseWritableFile` as far as I can tell, not in `PosixFileSystem::ReuseWritableFile` either.

@metaspace : Opening the file without O_APPEND will make writes to overwrite from file beginning, right? see the file open flags set in the posix ReuseWritableFile.

yhr:
And sorry, I did not look at the right piece of code when checking what we are doing now in PosixWritableFile.

The current implementation IS broken. But I do believe that the right thing to do here is to delete the old file if it exists and open a new file for writing using the new file name. We can't do in-place overwrites (as it looks like the posix implementation does)

aravind-wdc:
> But I do believe that the right thing to do here is to delete the old file if it exists and open a new file for writing using the new file name. We can't do in-place overwrites (as it looks like the posix implementation does)

Yes, we do not have to rename and reopen the file.
Changed the behavior to as explained above.

yhr:
Fixed in https://github.com/westerndigitalcorporation/zenfs/pull/129

PR #94, opened 2021-11-23
fs, util: Introducing semi automatic versioning for ZenFS.

MaisenbacherD:
When building ZenFS through RocksDB the ZenFS version is updated by
its Makefile and contains the output of `git describe --dirty`.
This means that the version depends on the latest git tag and if
preceding commits exits the latest commit's short hash
(see (documentation)[https://git-scm.com/docs/git-describe]).

The version is regenerated if the git HEAD changed at build time.

Major, minor, patch, the short commit hash and a dirty flag is
persisted within the superblock. A new util command `fs-info` was
added that outputs the contents of the current superblock which
includes the ZenFS version information.

Suggested-by: Andreas Hindborg <andreas.hindborg@wdc.com>

### DIFF
--- a/fs/fs_zenfs.h
+++ b/fs/fs_zenfs.h
@@ -13,6 +13,7 @@
 #include "rocksdb/env.h"
 #include "rocksdb/file_system.h"
 #include "rocksdb/status.h"
+#include "version.h"
 #include "zbd_zenfs.h"
 
 namespace ROCKSDB_NAMESPACE {
@@ -26,19 +27,20 @@ class Superblock {
   uint32_t magic_ = 0;
   char uuid_[37] = {0};
   uint32_t sequence_ = 0;
-  uint32_t version_ = 0;
+  uint32_t superblock_version_ = 0;
   uint32_t flags_ = 0;
   uint32_t block_size_ = 0; /* in bytes */
   uint32_t zone_size_ = 0;  /* in blocks */
   uint32_t nr_zones_ = 0;
   char aux_fs_path_[256] = {0};
   uint32_t finish_treshold_ = 0;
-  char reserved_[187] = {0};
+  char zenfs_version_[64]{0};
+  char reserved_[123] = {0};
 
  public:
   const uint32_t MAGIC = 0x5a454e46; /* ZENF */
   const uint32_t ENCODED_SIZE = 512;
-  const uint32_t CURRENT_VERSION = 1;
+  const uint32_t CURRENT_SUPERBLOCK_VERSION = 1;
   const uint32_t DEFAULT_FLAGS = 0;
 
   Superblock() {}
@@ -53,7 +55,7 @@ class Superblock {
                  sizeof(uuid_) - 1); /* make sure uuid is nullterminated */
     memcpy((void*)uuid_, uuid.c_str(), uuid_len);
     magic_ = MAGIC;
-    version_ = CURRENT_VERSION;
+    superblock_version_ = CURRENT_SUPERBLOCK_VERSION;
     flags_ = DEFAULT_FLAGS;
     finish_treshold_ = finish_threshold;
 
@@ -62,11 +64,16 @@ class Superblock {
     nr_zones_ = zbd->GetNrZones();
 
     strncpy(aux_fs_path_, aux_fs_path.c_str(), sizeof(aux_fs_path_) - 1);
+
+    std::string zenfs_version = ZENFS_VERSION;
+    strncpy(zenfs_version_, zenfs_version.c_str(), sizeof(zenfs_version_) - 1);
   }
 
   Status DecodeFrom(Slice* input);
   void EncodeTo(std::string* output);
   Status CompatibleWith(ZonedBlockDevice* zbd);
+
+  void GetReport(std::string* reportString);
 
   uint32_t GetSeq() { return sequence_; }
   std::string GetAuxFsPath() { return std::string(aux_fs_path_); }
@@ -188,6 +195,8 @@ class ZenFS : public FileSystemWrapper {
   }
 
   void EncodeJson(std::ostream& json_stream);
+
+  void ReportSuperblock(std::string* report) { superblock_->GetReport(report); }
 
   virtual IOStatus NewSequentialFile(const std::string& fname,
                                      const FileOptions& file_opts,
--- a/util/zenfs.cc
+++ b/util/zenfs.cc
@@ -539,6 +539,24 @@ int zenfs_tool_dump() {
   return 0;
 }
 
+int zenfs_tool_fsinfo() {
+  Status s;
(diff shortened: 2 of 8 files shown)


fs/fs_zenfs.h
```diff
@@ -56,12 +61,23 @@ class Superblock {
...
     strncpy(aux_fs_path_, aux_fs_path.c_str(), sizeof(aux_fs_path_) - 1);
+
+    major_zenfs_version_ =
+        static_cast<uint32_t>(std::stoul(ZENFS_VERSION_MAJOR));
+    minor_zenfs_version_ =
+        static_cast<uint32_t>(std::stoul(ZENFS_VERSION_MINOR));
+    patch_zenfs_version_ =
+        static_cast<uint32_t>(std::stoul(ZENFS_VERSION_PATCH));
+    strncpy(commit_zenfs_version_, std::string(ZENFS_VERSION_COMMIT).c_str(),
+            sizeof(commit_zenfs_version_) - 1);
```
metaspace, line 72: I don't think that you can actually be sure that `commit_zenfs_version_` is initialized to zero in all implementations. Which mean you don't have a null character at the end of the destination string. Also, no need to go through std::string as far as I can tell:
```suggestion
    strncpy(commit_zenfs_version_, ZENFS_VERSION_COMMIT,
        sizeof(commit_zenfs_version_));
     commit_zenfs_version_[8] = 0;
```

util/zenfs.cc
```diff
@@ -481,6 +481,28 @@ int zenfs_tool_dump() {
   return 0;
 }
 
+int zenfs_tool_fsinfo() {
+  Status s;
+  std::unique_ptr<ZonedBlockDevice> zbd = zbd_open(true, false);
+  if (!zbd) return 1;
+  ZonedBlockDevice *zbdRaw = zbd.get();
```
metaspace, line 488: It is sort of an anti pattern to take the raw value of a `unique_ptr` before passing a reference to a function. I see that this is the general pattern used, but we should probably change `zbd_open` to just return a raw pointer. We are not really using `unique_ptr` at all, we are just working around it.

zenfs.mk
```diff
@@ -1,3 +1,27 @@
...
+	$(eval ZENFS_VERSION_MAJOR := $(shell echo $(ZENFS_VERSION) | awk -F . '{ print substr($$1,2); }'))
+	$(eval ZENFS_VERSION_MINOR := $(shell echo $(ZENFS_VERSION) | awk -F . '{ print $$2 }'))
+	$(eval ZENFS_PATCH_AND_SUFFIX := $(shell echo $(ZENFS_VERSION) | awk -F . '{ print $$3 }'))
+	$(eval ZENFS_VERSION_PATCH := $(shell echo $(ZENFS_PATCH_AND_SUFFIX) | awk -F - '{ print $$1 }'))
+	$(eval ZENFS_VERSION_COMMIT := $(shell echo $(ZENFS_PATCH_AND_SUFFIX) | awk -F - '{ print substr($$3,2); }'))
+	$(eval ZENFS_VERSION_DIRTY := $(shell echo $(ZENFS_PATCH_AND_SUFFIX) | awk -F - 'NF == 4 { print "!" }'))
+	$(eval ZENFS_VERSION_COMMIT := $(ZENFS_VERSION_COMMIT)$(ZENFS_VERSION_DIRTY))
+	printf '#pragma once\n#define ZENFS_VERSION "%s"\n#define ZENFS_VERSION_MAJOR "%s"\n#define ZENFS_VERSION_MINOR "%s"\n#define ZENFS_VERSION_PATCH "%s"\n#define ZENFS_VERSION_COMMIT "%s"' $(ZENFS_VERSION) $(ZENFS_VERSION_MAJOR) $(ZENFS_VERSION_MINOR) $(ZENFS_VERSION_PATCH) $(ZENFS_VERSION_COMMIT) > $(ZENFS_ROOT_DIR)/fs/version.h
+	$(shell cd $(ZENFS_ROOT_DIR) && git update-index --assume-unchanged $(ZENFS_ROOT_DIR)/fs/version.h && git update-index --really-refresh > /dev/null)
```
metaspace, line 27: If possible, I think this shell kung fu should go into a separate shell script. It clutters the Makefile and makes it difficult to read.

metaspace (changes requested):
Nice work! I have a few comments :)

fs/fs_zenfs.h
```diff
@@ -20,19 +21,23 @@ class Superblock {
...
   uint32_t block_size_ = 0; /* in bytes */
   uint32_t zone_size_ = 0;  /* in blocks */
   uint32_t nr_zones_ = 0;
   char aux_fs_path_[256] = {0};
   uint32_t finish_treshold_ = 0;
-  char reserved_[187] = {0};
+  uint32_t major_zenfs_version_ = 0;
+  uint32_t minor_zenfs_version_ = 0;
+  uint32_t patch_zenfs_version_ = 0;
+  char commit_zenfs_version_[9] = {0}; /* 7 chars short sha + dirty indicator + nullterminator */
```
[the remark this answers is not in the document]
MaisenbacherD: Commit hash is saved as a number now and the dirty flag is also encoded within the commit_zenfs_version_

fs/fs_zenfs.h
```diff
@@ -56,12 +61,23 @@ class Superblock {
     nr_zones_ = zbd->GetNrZones();
 
     strncpy(aux_fs_path_, aux_fs_path.c_str(), sizeof(aux_fs_path_) - 1);
+
+    major_zenfs_version_ =
+        static_cast<uint32_t>(std::stoul(ZENFS_VERSION_MAJOR));
```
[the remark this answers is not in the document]
MaisenbacherD: Makes sense. I adjusted this

fs/fs_zenfs.h
MaisenbacherD: This is a non issue now because `commit_zenfs_version_` is now a `uint32_t` :)

util/zenfs.cc
MaisenbacherD: Agree with you, however this is a non issue for this PR now because `zbdRaw` was not in use, so I removed it.

zenfs.mk
MaisenbacherD: Separated the parsing into `generate-version.sh` and improved dependencies for building `version.h`

MaisenbacherD:
`version.h` depends on every file in this repo.

MaisenbacherD:
Improved the way how version.h is generated, which also fixes the build on a clean setup.

MaisenbacherD:
Fixed Issue reported by @yhr where `git describe` returns a commit slug length !=7

MaisenbacherD:
As discussed: We are now persisting the `git describe --abbrev=7 --dirty` output directly in the superblock rather than parsing the individual version components.

yhr (approved):

metaspace (approved):

PR #96, opened 2021-11-24
fs: Correct impl of `ZenFS::DeleteFile` when target does not exist

metaspace:
Return error when trying to delete non existing file, to align with
`PosixFileSystem::DeleteFile()`:

```c++
    if (unlink(fname.c_str()) != 0) {
      result = IOError("while unlink() file", fname, errno);
    }
```

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -621,34 +621,43 @@ IOStatus ZenFS::GetFileSize(const std::string& f, const IOOptions& options,
   return s;
 }
 
-IOStatus ZenFS::RenameFile(const std::string& f, const std::string& t,
+IOStatus ZenFS::RenameFile(const std::string& source_path,
+                           const std::string& dest_path,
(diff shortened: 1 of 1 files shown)


fs/fs_zenfs.cc
```diff
@@ -417,6 +417,8 @@ IOStatus ZenFS::DeleteFile(std::string fname) {
     } else {
       zoneFile.reset();
     }
+  } else {
+    s = IOStatus::NotFound("ZenfFS::DeleteFile(): File not found");
```
aravind-wdc, line 412: Typo: ZenfFS -> ZenFS.
Could we log the filename which was not found, will be useful in debugging issues.

metaspace: Thanks @aravind-wdc. Pushed a fix.

yhr (approved):
Looks good to me

yhr:
Please submit separate PRs for fixups (like the locking patch in this pr) and features / bugfixes next time.

PR #100, opened 2021-11-30
add Metrics patch

sagitrs:
I added ZenFSMetrics to the for_v2 branch.

ZenFSMetrics is used to monitor some of the variables on ZenFS in real time.
In addition to this, two other classes have been added to metrics.h.
NoZenFSMetrics is an empty implementation of ZenFSMetrics and is used by default.
ZenFSMetricsLatencyGuard is used to count the execution time of a function, 
and it reports the time between construction and destruction to Metrics.


skyzh:
Looks like your PR includes two commits. You may rebase and remove the 413b78b.

skyzh:
You should select "v2" as the PR base branch. Please click on "Edit" button and make the switch.

Issue #65, opened 2021-10-14
Unspected space amplification on rocksdb with zenfs


BaronStack:
## TEST
CPU: Intel(R) Xeon(R) Gold 6330 CPU
MEM: 125G
DISK: WZS4C8T1TDSP303 * 2T
Rocksdb: 6.25
Zenfs: master
max_active_zones: 14
max_open_zones: 14
chunk_sectors: 4194304
chunk_sectors-size: 2M

## Problem
Two much amplification after fillrandom and overwrite.

test_bench.sh
```shell
DEV=$1
 ./db_bench \
     --fs_uri=zenfs://dev:$DEV \
     --benchmarks=fillrandom,stats \
     -statistics \
     --num=3000000000 \
     --threads=32 \
     --db=./db \
     --wal_dir=./db \
     -report_interval_seconds=1 \
     -stats_dump_period_sec=5 \
     --duration=3000 \
     --key_size=16 \
     --value_size=128 \
     --max_write_buffer_number=16 \
     -max_background_compactions=32 \
     -max_background_flushes=16 \
     -subcompactions=8 \
     -compression_type=none \

 cp zns-log/db/LOG test-log/fillrandom.log
 sleep 10

./db_bench \
     --fs_uri=zenfs://dev:$DEV \
     --benchmarks=overwrite,stats \
     -statistics \
     --num=3000000000 \
     --threads=32 \
     --db=./db \
     --wal_dir=./db \
     -report_interval_seconds=1 \
     -stats_dump_period_sec=5 \
     --duration=1800 \
     --key_size=16 \
     --value_size=128 \
     --use_existing_db=1 \
     --use_existing_keys=1 \
     --max_write_buffer_number=16 \
     -max_background_compactions=32 \
     -max_background_flushes=16 \
     -subcompactions=8 \
     -compression_type=none \
```

After finished:
```shell
# ./plugin/zenfs/util/zenfs df --zbd=nvme2n1
Free: 140293 MB
Used: 54189 MB
Reclaimable: 776863 MB
Space amplification: 1433%
```
We can see that there ara more than 770G space need reclaim, and actual space occupy is just 54G.

Any suggestion to solve the problem?

skyzh:
There is a dump tool which could help analyze the on-disk layout of your RocksDB instance. You may have a try and upload logs. Meanwhile, would you please take a look at RocksDB log, better to upload it? Is there any warning that caused such space amplification?

BaronStack:
[...]

PR #103, opened 2021-12-03
fs: A patch adding GetSnapshot() into zenfs.

sagitrs:
These are snapshot classes to capture real-time information from ZenFS
for use by the upper layer algorithms. The three snapshots will capture
information from Zone, ZoneFile, and ZoneExtent respectively.
There are new methods in ZoneBlockDevice and ZenFS to get snapshots of ZenFS.

_There is a new Patch being used to support more flexible Snapshot fetching, 
passing in an Option to get a subset of the ZenFS parameters instead of all of them._
Removed (may be a future PR:)

Signed-off-by: Liu Ruicheng <sagitrs67@gmail.com>

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -20,6 +20,7 @@
 
 #include "metrics_sample.h"
 #include "rocksdb/utilities/object_registry.h"
+#include "snapshot.h"
 #include "util/coding.h"
 #include "util/crc32c.h"
 
@@ -1169,6 +1170,17 @@ std::map<std::string, std::string> ListZenFileSystems() {
   }
 
   return zenFileSystems;
+}
+
+void ZenFS::GetZoneSnapshot(std::vector<ZoneSnapshot>& zones) {
+  zbd_->GetZoneSnapshot(zones);
+}
+
+void ZenFS::GetZoneFileSnapshot(std::vector<ZoneFileSnapshot>& zone_files) {
+  std::lock_guard<std::mutex> file_lock(files_mtx_);
+  for (auto& file_it : files_) {
+    zone_files.emplace_back(*file_it.second);
+  }
 }
 
 extern "C" FactoryFunc<FileSystem> zenfs_filesystem_reg;
--- a/fs/snapshot.h
+++ b/fs/snapshot.h
@@ -1,0 +1,78 @@
+// Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
+// Copyright (c) 2019-present, Western Digital Corporation
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#pragma once
+
+#include <string>
+#include <vector>
+#include "io_zenfs.h"
+#include "zbd_zenfs.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// These are three Snapshot classes to capture real-time information from ZenFS
+// for use by the upper layer algorithms. The three Snapshots will capture
+// information from Zone, ZoneFile, and ZoneExtent respectively. If you plan to
+// modify the variables of these three classes, please make sure that they have
+// a public interface for copying and that the interface of the Snapshot classes
+// are still logically correct after modification.
+
+class ZoneSnapshot {
+ private:
+  uint64_t start_;
+  uint64_t capacity_;
+  uint64_t max_capacity_;
+  uint64_t wp_;
+
+ public:
+  ZoneSnapshot(const Zone& zone)
+      : start_(zone.start_),
+        capacity_(zone.capacity_),
+        max_capacity_(zone.max_capacity_),
+        wp_(zone.wp_) {}
+
+  uint64_t ID() const { return start_; }
+  uint64_t RemainingCapacity() const { return capacity_; }
+  uint64_t MaxCapacity() const { return max_capacity_; }
+  uint64_t StartPosition() const { return start_; }
+  uint64_t WritePosition() const { return wp_; }
+};
+
+struct ZoneExtentSnapshot {
+ private:
+  uint64_t start_;
+  uint64_t length_;
+  uint64_t zone_start_;
+
+ public:
+  ZoneExtentSnapshot(const ZoneExtent& extent)
+      : start_(extent.start_),
+        length_(extent.length_),
+        zone_start_(extent.zone_->start_) {}
+
+  uint64_t Start() const { return start_; }
+  uint64_t Length() const { return length_; }
+  uint64_t ZoneID() const { return zone_start_; }
+};
+
+struct ZoneFileSnapshot {
+ private:
+  uint64_t file_id_;
+  std::string filename_;
+  std::vector<ZoneExtentSnapshot> extent_;
+
+ public:
+  ZoneFileSnapshot(ZoneFile& file)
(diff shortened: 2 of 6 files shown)


metaspace:
Hi @sagitrs, thank you for the PR!

What is the intended use case for this patch? What code is the intended caller? Could you add an extra commit with and example of how this API is supposed to be used by upper layers?

fs/fs_zenfs.cc
```diff
@@ -1171,6 +1172,18 @@ std::map<std::string, std::string> ListZenFileSystems() {
   return zenFileSystems;
 }
 
+void ZenFS::GetZoneSnapshot(std::vector<ZoneSnapshot>& zones) {
+  zbd_->GetZonesSnapshot(zones);
+}
+
+void ZenFS::GetZoneFileSnapshot(std::vector<ZoneFileSnapshot>& zone_files) {
+  files_mtx_.lock();
```
MaisenbacherD, line 1180: For locking and unlocking `files_mtx_` we use this pattern:
`std::lock_guard<std::mutex> file_lock(files_mtx_);`

fs/fs_zenfs.cc
```diff
@@ -1171,6 +1172,18 @@ std::map<std::string, std::string> ListZenFileSystems() {
...
+  for (auto& file_it : files_) {
```
MaisenbacherD, line 1181: Since the original `files_` should not be modified we should use `auto const& file_it : files_` here.

skyzh:
@metaspace I believe this PR is the same as https://github.com/westerndigitalcorporation/zenfs/pull/43, and the use case is nearly the same.

fs/snapshot.h
```diff
@@ -0,0 +1,81 @@
...
+};
+
+struct ZoneFileSnapshot {
+ private:
+  uint64_t file_id_;
+  std::string filename_;
+  std::vector<ZoneExtentSnapshot> extent_;
+
+ public:
+  ZoneFileSnapshot(ZoneFile& file)
```
MaisenbacherD, line 70: Same here: We don't want the ZoneFile to be modified so we should use `const`

skyzh:
> In fact, this is mainly used to reduce the tremendous space amplification in our workload. For RocksDB on ZNS, this issue might have been well-resolved in https://github.com/westerndigitalcorporation/zenfs/issues/65. But in our use case TerarkDB, space amplification is still a big problem.
> 
> * TerarkDB implements key-value separation, where values from SSTs of different files might be now in a single blob file, which means that lifetime of each file is now meaningless.
> * And this might cause that a zone only contains a small file, due to lack of GC in ZenFS.
> 
> Therefore, we added a stats API to ZenFS, where TerarkDB could read from, and decide on which files to compact, therefore moving those small files away and free up zone space.

> In a nutshell, TerarkDB will select files to compact based on this stats API, so as to reduce space amplification. See https://github.com/bytedance/terarkdb/pull/87 for the TerarkDB side's implementation. In our case, this significantly reduced space amplification reported by ZenFS tools, 300% -> 100% or even less.

fs/snapshot.h
```diff
@@ -0,0 +1,81 @@
...
+      : file_id_(file.GetID()), filename_(file.GetFilename()), extent_() {
+    for (ZoneExtent*& extent : file.GetExtents()) extent_.emplace_back(*extent);
```
MaisenbacherD, line 72: Is `ZoneExtent*&` having any advantage over `const ZoneExtent*`?

yhr (approved):
Looks allright to me, we might improve upon this in the future, but this is an ok starting point.

yhr:
I can fix up the lint warning and push this manually

fs/zbd_zenfs.h
```diff
@@ -152,6 +153,8 @@ class ZonedBlockDevice {
 
   std::shared_ptr<ZenFSMetrics> GetMetrics() { return metrics_; }
 
+  void GetZonesSnapshot(std::vector<ZoneSnapshot> &snapshot);
```
MaisenbacherD, line 156: Maybe its just me: Renaming this to `GetZoneSnapshots` sounds a little better.

fs/snapshot.h
```diff
@@ -0,0 +1,81 @@
...
+  uint64_t wp_;
+
+ public:
+  ZoneSnapshot(const Zone& zone)
+      : start_(zone.start_),
+        capacity_(zone.capacity_),
+        max_capacity_(zone.max_capacity_),
+        wp_(zone.wp_) {}
+
+ public:
```
MaisenbacherD, line 37: Should we remove duplicate `public` access specifier of a class/struct?

fs/fs_zenfs.cc
sagitrs: Yes, but I found out when I used ZoneFile::GetID() that it is not a const function, and I am not sure if there is a reason for this design. The same problem happens elsewhere, and I think maybe this would be better as a separate PR.

fs/zbd_zenfs.h
sagitrs: Fixed and yes, I also think this is a typo.

fs/snapshot.h
sagitrs: Fixed, this is not related to the previously mentioned issue, it is a separate bug.

fs/snapshot.h
sagitrs: Yes, it is somewhat redundant since only public functions are available.

fs/fs_zenfs.cc
sagitrs: Fixed.

fs/snapshot.h
sagitrs: The same problem was encountered with the former.

sagitrs:
@metaspace As @skyzh said, we have an algorithm in TerarkDB that needs to use the ZenFS snapshot information.

fs/fs_zenfs.cc
MaisenbacherD: Okay makes sense :)

sagitrs:
I forgot to add sign-off for @skyzh , I will add it in rebase.
I added an additional patch package to support freer data fetching, but it may still have some bugs to fix.
If this patch package is considered redundant, I will withdraw the change and submit this PR when we use this feature in the future.

yhr:
@sagitrs : Only add the code you know that you'll be using now. We're happy to take follow up pull requests.

MaisenbacherD:
Looks good to me now :)

sagitrs:
> @sagitrs : Only add the code you know that you'll be using now. We're happy to take follow up pull requests.
@yhr Thanks for the suggestion, I rolled back the patch and rebase it.

yhr:
Pushed manually after a couple of trivial fixups making the linter happy

yhr:
https://github.com/westerndigitalcorporation/zenfs/commit/7cad1d7cfa123a739f0d3364f9042120cc9985de

skyzh:
> I forgot to add sign-off for @skyzh , I will add it in rebase.

If you need to attribute others' contribution, you may need to include both the Sign-Off and the Co-Authored-By footers in the commit message.

e.g.
```
Signed-off-by: Name <email>
Co-Authored-By: Name <email>
```

> I added an additional patch package to support freer data fetching, but it may still have some bugs to fix.
> If this patch package is considered redundant, I will withdraw the change and submit this PR when we use this feature in the future.

One PR should only do a small change. I think the `GetStat` API or `GetSnapshot` API is enough for one PR. I believe we can contribute little by little.

PR #105, opened 2021-12-08
zenfs: change include header path to relative path

royguo:
Some variants of RocksDB (and older RocksDB version) may not have a `plugin` directory, so I am changing the include path a relative one.

### DIFF
--- a/util/zenfs.cc
+++ b/util/zenfs.cc
@@ -7,7 +7,6 @@
 #include <fcntl.h>
 #include <gflags/gflags.h>
 #include <rocksdb/file_system.h>
-#include <rocksdb/plugin/zenfs/fs/fs_zenfs.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 
@@ -18,6 +17,12 @@
 #include <memory>
 #include <sstream>
 #include <streambuf>
+
+#ifdef WITH_TERARKDB
+#include <fs/fs_zenfs.h>
+#else
+#include <rocksdb/plugin/zenfs/fs/fs_zenfs.h>
+#endif
 
 using GFLAGS_NAMESPACE::ParseCommandLineFlags;
 using GFLAGS_NAMESPACE::RegisterFlagValidator;


util/Makefile
```diff
@@ -7,8 +7,9 @@ CXX ?= g++
 
 CXXFLAGS = $(shell pkg-config --cflags rocksdb)
 LIBS = $(shell pkg-config --static --libs rocksdb)
+INC = $(shell dirname $(abspath $(dir $$PWD)))
```
MaisenbacherD, line 10:
```suggestion
INC = $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
```

MaisenbacherD: PWD depends on where you build from. :)

royguo: @MaisenbacherD Thanks, PTAL. It seems this Makefile would only be used inside the `util` directory to compile `zenfs.cc` manually. The CMake script is not affected by this Makefile file.

metaspace:
I think this is a problem best solved in the code that includes the ZenFS plugin. That code should install ZenFS headers, and the correct include path for finding ZenFS headers should be returned by `pkg-config --cflags rocksdb`.

If we build against local `fs_zenfs` we may get into problems where we are building against a header that does not mach the library that is linked in.

We could make the library to link against configurable, as in `pkg-config --cflags $(HOST_DB)` to link against other plugin hosts.

royguo:
@metaspace Understood. Current `CMakeLists` in both TerarkDB and RocksDB will find the header files automatically and build the `zenfs` tool without a problem.

The main purpose of this patch is to change the include path to a relative path (which is important to TerarkDB and has no harm for RocksDB).

@metaspace Do you think we can merge this patch and you can have further fixes in the further? The new Makefile just wanna to we don't break anything at the moment.

metaspace:
> @metaspace Understood. Current `CMakeLists` in both TerarkDB and RocksDB will find the header files automatically and build the `zenfs` tool without a problem.
> 
> The main purpose of this patch is to change the include path to a relative path (which is important to TerarkDB and has no harm for RocksDB).
> 
> @metaspace Do you think we can merge this patch and you can have further fixes in the further? The new Makefile just wanna to we don't break anything at the moment.

As I said, I do not think it is a good idea to include relative, as we risk including a header that does not match the linked library. I think this should be fixed in TerakDB. TerakDB should install `fs/fs_zenfs.h` somewhere, and `pkg-config` will report where that somewhere is.

If TerakDB does not install ZenFS headers, then that should be fixed. If TerakDB installs ZenFS headers to a different location, then we can add an `#ifdef` around the include, to include the right path when building for TerakDB.

@yhr what do you think?

royguo:
@metaspace TerarkDB (And older version of RocksDB) doesn't have a directory `plugin/xxxx` so even we install zenfs's header files, we cannot find this <rocksdb/plugin/zenfs/fs/fs_zenfs.h> file path.

In TerarkDB we do install all zenfs's header files and we add the `zenfs` root directory to our include path, that's why we can build zenfs.cc without a problem. I may have confused you by using the term `relative path`, the truth is we have an installed  path `zenfs/fs/fs_zenfs.fs`, not "rocksdb/plugin/zenfs/fs/fs_zenfs.h".

Do you have further idea to fix this for TerarkDB and older version RocksDB ? @metaspace @yhr

metaspace:
> @metaspace TerarkDB (And older version of RocksDB) doesn't have a directory `plugin/xxxx` so even we install zenfs's header files, we cannot find this <rocksdb/plugin/zenfs/fs/fs_zenfs.h> file path.
> 
> In TerarkDB we do install all zenfs's header files and we add the `zenfs` root directory to our include path, that's why we can build zenfs.cc without a problem. I may have confused you by using the term `relative path`, the truth is we have an installed path `zenfs/fs/fs_zenfs.fs`, not "rocksdb/plugin/zenfs/fs/fs_zenfs.h".
> 
> Do you have further ideas to fix this for TerarkDB and the older version RocksDB ? @metaspace @yhr

I think we could do something like
```c++
#ifdef TERAKDB
#include <terak-db-include-path-for-zenfs/fs/fs_zenfs.h>
#else
#include <rocksdb/plugin/zenfs/fs/fs_zenfs.h>
#endif
```
substitute `terak-db-include-path-for-zenfs/fs/fs_zenfs.h` for the right path to use when building against TerakDB.

royguo:
@metaspace Thanks, If define a new TERARKDB macro is acceptable, then it would be much easier

royguo:
@metaspace PLTAL, if this change works for you.

metaspace:
> @metaspace PLTAL, if this change works for you.

That looks good to me.

yhr:
Looks good to me, I can pull this in as a single commit once the smoke testing completes OK

PR #106, opened 2021-12-08
fs: Add support for getting ZBD status.

sagitrs:
Getting the state of the ZBD is different from others, for example the ZBD does not intentionally maintain a variable called free_space_, so we need to calculate the free space via a function GetFreeSpace().
    
Also, sometimes we would like to report these temporarily calculated values through metrics, and it is not a good idea to periodically calculate a variable in a specific function. So we added a ReportSnapshot interface to Metrics and left it up to the debugger to decide when to report them: the user can decide which variables to compute online via SnapshotOptions and then report them one by one via Metrics::ReportSnapshot().
    
We have added a sample related to this in metrics_example.h.
    
Signed-off-by: Liu Ruicheng <liuruicheng.222@bytedance.com>

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -282,9 +282,9 @@ IOStatus ZenFS::RollMetaZoneLocked() {
   Zone* new_meta_zone = nullptr;
   IOStatus s;
 
-  ZenFSMetricsLatencyGuard guard(zbd_->GetMetrics(), ZENFS_ROLL_LATENCY,
+  ZenFSMetricsLatencyGuard guard(zbd_->GetMetrics(), ZENFS_LABEL(ROLL, LATENCY),
                                  Env::Default());
-  zbd_->GetMetrics()->ReportQPS(ZENFS_ROLL_QPS, 1);
+  zbd_->GetMetrics()->ReportQPS(ZENFS_LABEL(ROLL, QPS), 1);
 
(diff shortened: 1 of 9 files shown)


sagitrs:
The reason for this rebase is that I missed the metrics_sample.h in the clang-format.

sagitrs:
In util/zenfs.cc I changed the inclue on line 10 due to the lack of a plugins directory in terarkdb, which would cause terarkdb to fail to compile.
We would like to fix this issue with PR if possible. If this causes a compilation failure I will withdraw this change.

util/zenfs.cc
```diff
@@ -7,7 +7,7 @@
 #include <fcntl.h>
 #include <gflags/gflags.h>
 #include <rocksdb/file_system.h>
-#include <rocksdb/plugin/zenfs/fs/fs_zenfs.h>
+#include "fs/fs_zenfs.h"
```
skyzh, line 10: It seems that this PR is based on https://github.com/westerndigitalcorporation/zenfs/pull/105 ? We need to merge #105 before proceeding.

sagitrs: Yes, I think we submitted conflicting changes.
The change has been withdrawn.

sagitrs:
The second patch was used to support IOType-based reporting information classification, and in addition, in this patch we fixed some naming irregularities due to historical reasons.
Since GetIOType is not supported yet, some variable fetch operations are left blank.

yhr (approved):
Looks good to me

PR #113, opened 2021-12-14
log: enable logger by default and add garbage stats log

royguo:
We need to log important information in production mode, so for Release
or RelWithDebInfo build type, we could set log level to INFO_LEVEL.

The garbage stats log helps us understand the garbage distribution of
the disk and also let us know whether the collaborative GC works as
expected.

The log looks like this:

```
2021/12/14-22:43:21.776104 7f944bdfe700 Zone Garbage Stats: [545 10 2 8 6 15 16 12 9 24 9 0 ]
2021/12/14-22:43:22.777050 7f944bdfe700 Zone Garbage Stats: [544 11 2 8 6 15 16 12 9 24 9 0 ]
2021/12/14-22:43:23.777990 7f944bdfe700 Zone Garbage Stats: [544 11 2 8 6 15 16 12 9 24 9 0 ]
2021/12/14-22:43:24.778996 7f944bdfe700 Zone Garbage Stats: [545 10 1 8 7 15 16 12 9 24 9 0 ]
2021/12/14-22:43:25.779879 7f944bdfe700 Zone Garbage Stats: [545 10 1 8 7 15 16 12 9 24 9 0 ]
2021/12/14-22:43:26.780911 7f944bdfe700 Zone Garbage Stats: [545 10 1 8 7 15 16 12 9 24 9 0 ]
2021/12/14-22:43:27.782006 7f944bdfe700 Zone Garbage Stats: [545 10 1 8 6 15 16 13 9 24 9 0 ]
2021/12/14-22:43:28.783024 7f944bdfe700 Zone Garbage Stats: [545 10 1 8 6 15 16 13 9 24 9 0 ]
2021/12/14-22:43:29.783971 7f944bdfe700 Zone Garbage Stats: [545 10 1 8 6 15 16 12 9 25 9 0 ]
2021/12/14-22:43:30.784890 7f944bdfe700 Zone Garbage Stats: [545 10 1 8 6 15 16 12 9 25 9 0 ]
2021/12/14-22:43:31.785847 7f944bdfe700 Zone Garbage Stats: [545 9 2 8 6 15 16 12 9 25 9 0 ]
2021/12/14-22:43:32.786764 7f944bdfe700 Zone Garbage Stats: [544 10 2 8 6 15 16 12 9 24 10 0 ]
2021/12/14-22:43:33.787730 7f944bdfe700 Zone Garbage Stats: [544 10 2 8 6 15 16 12 9 24 10 0 ]
2021/12/14-22:43:34.788693 7f944bdfe700 Zone Garbage Stats: [545 9 1 8 7 15 16 12 9 24 10 0 ]
2021/12/14-22:43:35.789895 7f944bdfe700 Zone Garbage Stats: [545 9 1 8 7 15 16 12 9 24 10 0 ]
```

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -1108,7 +1108,6 @@ std::map<std::string, Env::WriteLifeTimeHint> ZenFS::GetWriteLifeTimeHints() {
   return hint_map;
 }
 
-#ifndef NDEBUG
 static std::string GetLogFilename(std::string bdev) {
   std::ostringstream ss;
   time_t t = time(0);
@@ -1120,19 +1119,26 @@ static std::string GetLogFilename(std::string bdev) {
 
   return ss.str();
 }
-#endif
 
 Status NewZenFS(FileSystem** fs, const std::string& bdevname,
                 std::shared_ptr<ZenFSMetrics> metrics) {
   std::shared_ptr<Logger> logger;
   Status s;
 
-#ifndef NDEBUG
+  // TerarkDB needs to log important information in production while ZenFS
+  // doesn't (currently).
+  //
+  // TODO(guokuankuan@bytedance.com) We need to figure out how to reuse
+  // RocksDB's logger in the future.
+#if defined(NDEBUG) || defined(WITH_TERARKDB)
   s = Env::Default()->NewLogger(GetLogFilename(bdevname), &logger);
   if (!s.ok()) {
     fprintf(stderr, "ZenFS: Could not create logger");
   } else {
     logger->SetInfoLogLevel(DEBUG_LEVEL);
+#ifdef WITH_TERARKDB
+    logger->SetInfoLogLevel(INFO_LEVEL);
+#endif
   }
 #endif
 
@@ -1227,8 +1233,11 @@ void ZenFS::GetZenFSSnapshot(ZenFSSnapshot& snapshot,
     for (auto& file_it : files_)
       snapshot.zone_files_.emplace_back(*file_it.second, options);
   }
-  if (options.trigger_report_)
+  if (options.trigger_report_) {
     zbd_->GetMetrics()->ReportSnapshot(snapshot, options);
+  }
+
+  zbd_->LogGarbageInfo();
 }
 
 extern "C" FactoryFunc<FileSystem> zenfs_filesystem_reg;
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -431,6 +431,36 @@ void ZonedBlockDevice::LogZoneUsage() {
   }
 }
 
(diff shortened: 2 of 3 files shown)


royguo:
@metaspace You are right, we could place this log in a better place, I prefer to put this into a background worker since it does take some time to finish the calculation.

I am planning to add a `Background Worker` after @yhr's rebased branch was merged into master, then we can move `LogGarbageStats` to the background worker (and other heavy-weight functions, e.g. metadata zone allocation)

yhr:
@royguo : LogZoneStat was helpful when developing zenfs, but I think we could replace it with this kind of stats(that would be generated periodically) and enable it to be piped to the RocksDB/TerarkDB LOG file.

fs/fs_zenfs.cc
```diff
@@ -1120,21 +1119,22 @@ static std::string GetLogFilename(std::string bdev) {
...
-#ifndef NDEBUG
   s = Env::Default()->NewLogger(GetLogFilename(bdevname), &logger);
   if (!s.ok()) {
     fprintf(stderr, "ZenFS: Could not create logger");
   } else {
+#ifndef NDEBUG
     logger->SetInfoLogLevel(DEBUG_LEVEL);
-  }
+#else
+    logger->SetInfoLogLevel(INFO_LEVEL);
```
yhr, line 1135: I do not want to have a log being recorded in /tmp/ for production builds per default (this is undocumented, and the log messages are essentially debug information) 

Could you make this behavior specific to terarkdb? (make WITH_TERARKDB being set as a prerequisite for setting up the logger and configuring the level)

fs/fs_zenfs.cc
```diff
@@ -1227,8 +1227,12 @@ void ZenFS::GetZenFSSnapshot(ZenFSSnapshot& snapshot,
     for (auto& file_it : files_)
       snapshot.zone_files_.emplace_back(*file_it.second, options);
   }
-  if (options.trigger_report_)
+  if (options.trigger_report_) {
     zbd_->GetMetrics()->ReportSnapshot(snapshot, options);
+  }
+
+  // Log zone garbage info
```
yhr, line 1234: No need for the comment, the method name is very informative :)

yhr (changes requested):
If this behavior is made specific to terarkdb, I see no issue with it. See my comments.

It would be nice to be able to enable logging on production builds as well, we could do this with a mkfs option(specifying log level). We should also then move the log files to the aux dir. But that is a separate task.

fs/fs_zenfs.cc
royguo: Sounds good to me, we still need to find a way to re-use RocksDB's logger in the future.

yhr:
Looks good to me, fixed the lint issues and pulled it in via  https://github.com/westerndigitalcorporation/zenfs/pull/116

PR #117, opened 2021-12-20
Implement inline extents

yhr:
This PR reduces write tail latency and increases throughput for use cases that require synced writes (db_bench --sync=1 ..)  when all written key/values must be persisted directly and readable after a potenial crash.

The trick is to store the active extent start that we are writing to for each file and use the write pointer to recover the data in case of a crash. For non-block-size aligned writes, like the WAL, a sparse format is introduced where the length of each non-aligned segment is prefixed to the data, so each write contains the valid data length.

With these changes, we only need to write to the metadata zone when we a) allocate a new extent b) close the file.

```
 fs: implement file recovery based on active extent information
    
    We can now recover data from files that has not been fsynced
    by busing active extent start and the wp in the active extent
    zone's wp.
    
    For sparse files(that has segment lenght in line
    with the data stored) we need to go through each individual
    segment and re-create all non-synced extents.
    
 fs: Split buffered and non-buffered appends
    
    Make buffered writes sparse.
    
    Now that we store the extent lengths in line with data
    for every buffer flush, we don't have to write metadata records
    when asked to do a data sync.
    
 fs: refactor ZoneFile::Append
    
 fs: store active extent start
    
    If we store the current extent start and there is a crash
    we know that this file wrote the data starting from
    active extent start up to the zone write pointer.
    
    We don't need to store the active zone as we can look it up
    from extent_start_
    
```

This PR requires a bit of extra work before we can release it. The following issue has been created to cover that:

https://github.com/westerndigitalcorporation/zenfs/issues/118

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -254,6 +254,19 @@ ZenFS::~ZenFS() {
   delete zbd_;
 }
 
+IOStatus ZenFS::Repair() {
+  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
+  for (it = files_.begin(); it != files_.end(); it++) {
+    std::shared_ptr<ZoneFile> zFile = it->second;
+    if (zFile->HasActiveExtent()) {
+      IOStatus s = zFile->Recover();
+      if (!s.ok()) return s;
+    }
+  }
+
+  return IOStatus::OK();
+}
+
 void ZenFS::LogFiles() {
   std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   uint64_t total_size = 0;
@@ -1026,6 +1039,9 @@ Status ZenFS::Mount(bool readonly) {
       valid_logs[i].reset();
     }
   }
+
+  s = Repair();
+  if (!s.ok()) return s;
 
   if (readonly) {
     Info(logger_, "Mounting READ ONLY");
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -393,7 +393,7 @@ IOStatus ZenFS::PersistRecord(std::string record) {
   return s;
 }
 
-IOStatus ZenFS::SyncFileMetadata(std::shared_ptr<ZoneFile> zoneFile) {
+IOStatus ZenFS::SyncFileMetadata(ZoneFile* zoneFile) {
   std::string fileRecord;
   std::string output;
   IOStatus s;
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -263,8 +263,9 @@ void ZenFS::LogFiles() {
     std::shared_ptr<ZoneFile> zFile = it->second;
     std::vector<ZoneExtent*> extents = zFile->GetExtents();
 
-    Info(logger_, "    %-45s sz: %lu lh: %d", it->first.c_str(),
-         zFile->GetFileSize(), zFile->GetWriteLifeTimeHint());
+    Info(logger_, "    %-45s sz: %lu lh: %d sparse: %u", it->first.c_str(),
+         zFile->GetFileSize(), zFile->GetWriteLifeTimeHint(),
+         zFile->IsSparse());
     for (unsigned int i = 0; i < extents.size(); i++) {
       ZoneExtent* extent = extents[i];
       Info(logger_, "          Extent %u {start=0x%lx, zone=%u, len=%u} ", i,
@@ -523,6 +524,8 @@ IOStatus ZenFS::NewWritableFile(const std::string& fname,
   } else {
     zoneFile->SetIOType(IOType::kUnknown);
   }
+
+  zoneFile->SetSparse(!file_opts.use_direct_writes);
 
   /* Persist the creation of the file */
   s = SyncFileMetadata(zoneFile);
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -60,6 +60,7 @@ enum ZoneFileTag : uint32_t {
   kExtent = 5,
   kModificationTime = 6,
   kActiveExtentStart = 7,
+  kIsSparse = 8,
 };
 
 void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
@@ -93,6 +94,10 @@ void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
    * from extent_start_ */
   PutFixed32(output, kActiveExtentStart);
   PutFixed64(output, extent_start_);
+
+  if (is_sparse_) {
+    PutFixed32(output, kIsSparse);
+  }
 }
 
 void ZoneFile::EncodeJson(std::ostream& json_stream) {
@@ -173,6 +178,9 @@ Status ZoneFile::DecodeFrom(Slice* input) {
           return Status::Corruption("ZoneFile", "Active extent start");
         extent_start_ = es;
         break;
+      case kIsSparse:
+        is_sparse_ = true;
+        break;
       default:
         return Status::Corruption("ZoneFile", "Unexpected tag");
     }
@@ -198,7 +206,8 @@ Status ZoneFile::MergeUpdate(std::shared_ptr<ZoneFile> update) {
     zone->used_capacity_ += extent->length_;
     extents_.push_back(new ZoneExtent(extent->start_, extent->length_, zone));
   }
-
+  extent_start_ = update->GetExtentStart();
+  is_sparse_ = update->IsSparse();
   MetadataSynced();
 
   return Status::OK();
@@ -243,8 +252,8 @@ ZoneFile::~ZoneFile() {
 
 IOStatus ZoneFile::CloseWR() {
   IOStatus s = CloseActiveZone();
+  extent_start_ = NO_EXTENT;
   open_for_wr_ = false;
-  metadata_writer_ = NULL;
   return s;
 }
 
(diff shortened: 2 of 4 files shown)


yhr:
Ran a full baseline performance test to check that I didnt break anything while rebasing and cleaning up the patch set, completed ok with expected results. Compile and smoke tested in terarkdb as well, all ok.

fs/io_zenfs.cc
```diff
@@ -415,8 +420,65 @@ IOStatus ZoneFile::AllocateNewZone() {
...
+
+  while (left) {
+    wr_size = left + ZoneFile::SPARSE_HEADER_SIZE;
+    if (wr_size > active_zone_->capacity_) wr_size = active_zone_->capacity_;
+
+    /* Pad to the next block boundary if needed */
+    uint32_t align = wr_size % block_sz;
+    uint32_t pad_sz = 0;
+
+    align = wr_size % block_sz;
```
metaspace, line 444: Double assigned align

fs/io_zenfs.cc
```diff
@@ -415,8 +420,65 @@ IOStatus ZoneFile::AllocateNewZone() {
...
+    if (align) pad_sz = block_sz - align;
+    if (pad_sz) memset(sparse_buffer + wr_size, 0x0, pad_sz);
```
metaspace, line 446: Is this not in risk of writing beyond `sparse_buffer` ?

fs/io_zenfs.cc
```diff
@@ -415,8 +420,65 @@ IOStatus ZoneFile::AllocateNewZone() {
...
+
+    if (active_zone_->capacity_ == 0) {
+      s = CloseActiveZone();
+      if (!s.ok()) {
+        return s;
+      }
+      if (left) {
+        memcpy((void*)(sparse_buffer + ZoneFile::SPARSE_HEADER_SIZE),
+               (void*)(sparse_buffer + wr_size), left);
+      }
```
metaspace, line 471: Why not move the pointer instead of the data?

fs/io_zenfs.cc
```diff
@@ -481,25 +542,35 @@ ZonedWritableFile::ZonedWritableFile(ZonedBlockDevice* zbd, bool _buffered,
...
+    size_t sparse_buffer_sz;
+
+    sparse_buffer_sz =
+        1024 * 1024 + block_sz; /* one extra block size for padding */
+    int ret = posix_memalign((void**)&sparse_buffer, sysconf(_SC_PAGESIZE),
+                             sparse_buffer_sz);
+
+    if (ret) sparse_buffer = nullptr;
+
+    assert(sparse_buffer != nullptr);
```
metaspace, line 654: Should we not return an error here?

fs/io_zenfs.cc
```diff
@@ -389,25 +398,32 @@ void ZoneFile::PushExtent() {
...
+  Zone* zone;
+  IOStatus s = zbd_->AllocateIOZone(lifetime_, io_type_, &zone);
+
+  if (!s.ok()) return s;
+  if (!zone) {
+    return IOStatus::NoSpace("Zone allocation failure\n");
+  }
+  SetActiveZone(zone);
+  extent_start_ = active_zone_->wp_;
+  extent_filepos_ = fileSize;
```
MaisenbacherD, line 416: Out of curiosity: Is there a reason why it is `fileSize` rather than `file_size_`?

fs/io_zenfs.cc
```diff
@@ -547,32 +633,30 @@ IOStatus ZonedWritableFile::Flush(const IOOptions& /*options*/,
...
-  return zoneFile_->CloseWR();
+IOStatus ZonedWritableFile::Close(const IOOptions& /*options*/,
+                                  IODebugContext* /*dbg*/) {
+  IOStatus s = DataSync();
+  if (!s.ok()) return s;
+
+  s = zoneFile_->CloseWR();
+  if (!s.ok()) return s;
+
+  return zoneFile_->PersistMetadata();
```
metaspace, line 745: Do we need this change? Seems like it does the same as calling `Fsync` but with more lines of code.

metaspace (changes requested):
Posted a few comments. Otherwise looks good!

I would in general recommend that patches be split into smaller atomic units. Don't combine refactoring with functional changes. It makes the patches difficult to review, and it takes much more time and energy than it should.

How was the recovery tested?

fs/io_zenfs.cc
metaspace: Just a typo I think. It should be file_size_. I bet there are tools that can lint these kinds of deviations from code style. You know, in Rust the compiler will issue a warning if you camelCase struct members :rofl:

fs/io_zenfs.cc
```diff
@@ -515,6 +515,100 @@ IOStatus ZoneFile::Append(void* data, int data_size) {
...
+      break;
+    }
+    recovered_segments++;
+
+    zone->used_capacity_ += extent_length;
+    extents_.push_back(new ZoneExtent(next_extent_start + SPARSE_HEADER_SIZE,
+                                      extent_length, zone));
+
+    uint64_t extent_blocks = 1 + extent_length / block_sz;
+    next_extent_start += extent_blocks * block_sz;
```
MaisenbacherD, line 555: I think it would be easier to understand like that:
```suggestion
    next_extent_start += extent_length + block_sz;
```

MaisenbacherD:
Also had a few comments, otherwise looks good to me :)

fs/io_zenfs.cc
yhr: fixed

fs/io_zenfs.cc
yhr: No, we allocate an extra block for padding.

fs/io_zenfs.cc
yhr: Because of alignment reasons, The data has to start on a page boundary to issue direct io.

fs/io_zenfs.cc
metaspace: We could add a comment here I think

fs/io_zenfs.cc
yhr: Yep, I've created https://github.com/westerndigitalcorporation/zenfs/issues/119 to fix that.

fs/io_zenfs.cc
yhr: For historical/no good reason :) Lets fix this in a follow up-patch

fs/io_zenfs.cc
yhr: It is needed, as we need to do a CloseWR() before PersistMetadata()

metaspace: Why is sync->persist->closeWR not good ?

fs/fs_zenfs.cc
```diff
@@ -254,6 +254,19 @@ ZenFS::~ZenFS() {
   delete zbd_;
 }
 
+IOStatus ZenFS::Repair() {
+  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
```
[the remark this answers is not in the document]
yhr: We don't need to lock as this is only done during mount, and there is no concurrency. RecoverFrom does not lock either.

fs/io_zenfs.cc
```diff
@@ -515,6 +515,100 @@ IOStatus ZoneFile::Append(void* data, int data_size) {
...
+  /* Figure out which zone we were writing to */
+  Zone* zone = zbd_->GetIOZone(extent_start_);
+
+  if (zone == nullptr) {
+    return IOStatus::IOError(
+        "Could not find zone for extent start while recovering");
+  }
+
+  if (zone->wp_ < extent_start_) {
+    return IOStatus::IOError("Zone wp bigger than active extent start");
```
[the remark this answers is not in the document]
yhr: fixed

fs/io_zenfs.cc
yhr: Hmm , the suggested change is not equivalent in behavior. The next extent starts on the next block boundary.

yhr:
> Posted a few comments. Otherwise looks good!
> 
> I would in general recommend that patches be split into smaller atomic units. Don't combine refactoring with functional changes. It makes the patches difficult to review, and it takes much more time and energy than it should.
> 
> How was the recovery tested?

I've tested the recovery by crashing db_bench and making sure that the database could be opened and recovered to the expected state afterwards. . Rocksdb does integrity checking on nearly everything it reads, so corruptions would have been caught. I made sure my testing uncovered actual corruptions when Repair() did not work.

That said, I believe we should do more systematic crash testing before the next release.

fs/io_zenfs.cc
yhr: We need to set extent_start_ = NO_EXTENT before persisting metadata

fs/fs_zenfs.cc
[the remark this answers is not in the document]
MaisenbacherD: Yep. This suggestion was meant as a precaution in case `Repair` is used somewhere else than for mounting at some point. :)

fs/io_zenfs.cc
MaisenbacherD: Created an issue for that https://github.com/westerndigitalcorporation/zenfs/issues/120

fs/io_zenfs.cc
MaisenbacherD: Maybe I am missing something here, but the suggestion should be equivalent (given that `extent_length` is a multiple of `block_sz`):
`(1 + extent_length / block_sz) * block_sz == block_sz * 1 + block_sz * extent_length / block_sz == block_sz + extent_length`

MaisenbacherD: Oh wait, I think that I got your point. `extent_length` is not block aligned (thus not a multiple of `block_sz`) and therefore you are right.

fs/io_zenfs.cc
yhr: That is a jolly good idea

yhr:
Adressed the review comments and made notes of follow up work (crash testing in the CI, fix camel-calse fileSize)

PR #121, opened 2021-12-23
zbd: Allow FinishCheapestZone to fail without break the zone allocation

royguo:
We could run into a situation where all zones are empty or full except busy ones.

In such cases, the `FinishCheapest` could return an error of failing to find a victim, but I believe we should keep trying until we get an active zone.

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -220,6 +220,9 @@ IOStatus ZonedBlockDevice::Open(bool readonly, bool exclusive) {
   Status s;
   uint64_t i = 0;
   uint64_t m = 0;
+  // Reserve one zone for metazone allocation, may add more for other
+  // tasks in the future.
+  int reserved_zones = 1;
   int ret;
 
   if (!readonly && !exclusive)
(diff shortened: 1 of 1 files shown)


fs/zbd_zenfs.cc
```diff
@@ -773,9 +776,10 @@ IOStatus ZonedBlockDevice::AllocateIOZone(Env::WriteLifeTimeHint file_lifetime,
       /* We have to make sure we can open an empty zone */
       while (!got_token && !GetActiveIOZoneTokenIfAvailable()) {
         s = FinishCheapestIOZone();
+        // Note that we may run into a situation that all zones are either empty
+        // or full except busy ones, so we have to keep trying.
         if (!s.ok()) {
-          PutOpenIOZoneToken();
-          return s;
+          Info(logger_, "Finish cheapest zone failed, keep trying...");
```
metaspace, line 782: We should probably find a way to check if we fail because all zones are empty or full, or if we are returning because of another error.

metaspace (changes requested):
Thank you for the PR, I think you are right that we may fail if all zones are empty or full. But we need to only retry if that was the case, not for other errors.

fs/zbd_zenfs.cc
royguo: your are right, maybe let the FinishCheapXx function return success when no zone is ready to finish is better.

MaisenbacherD (approved):
Looks good to me

yhr:
Looks good to me, ran the linter manually and that passed without any errors, so pulling this in.

Issue #72, opened 2021-10-28
db_stress failure


aravind-wdc:
db_stress fails with following error:

sudo ./db_stress --ops_per_thread=1000 --reopen=5 --fs_uri=zenfs://dev:nvme2n2

WARNING: prefix_size is non-zero but memtablerep != prefix_hash
DB path: [rocksdbtest/dbstress]
(Re-)verified 0 unique IDs
Error opening unique id file for append: IO error: No such file or directory: While open a file for appending: rocksdbtest/dbstress/.unique_ids: No such file or directory
Choosing random keys with no overwrite
Creating 2621440 locks
Starting continuous_verification_thread
2021/10/28-11:47:57  Initializing worker threads
2021/10/28-11:47:57  Starting database operations
2021/10/28-11:47:57  Reopening database for the 1th time
WARNING: prefix_size is non-zero but memtablerep != prefix_hash
DB path: [rocksdbtest/dbstress]
Segmentation fault

Probably due to https://github.com/facebook/rocksdb/pull/8990
and 0534393fc889e6dc44eb69d89fd2385082b4bddb

yhr:
fixed upstream, closing

PR #125, opened 2022-01-02
fs: fix register pattern entry

SheldonZhong:
RocksDB upstream no longer supports Regex. See the PR [here](https://github.com/facebook/rocksdb/pull/9264).

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -922,8 +922,6 @@ Status ZenFS::RecoverFrom(ZenMetaLog* log) {
     return Status::NotFound("ZenFS", "No snapshot found");
 }
 
-#define ZENV_URI_PATTERN "zenfs://"
-
 /* Mount the filesystem by recovering form the latest valid metadata zone */
 Status ZenFS::Mount(bool readonly) {
   std::vector<Zone*> metazones = zbd_->GetMetaZones();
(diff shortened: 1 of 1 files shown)


MaisenbacherD (approved):
Looks good to me

metaspace:
I get the following error when building:

```text
plugin/zenfs/fs/fs_zenfs.cc:1281:24: error: 'PatternEntry' is not a member of 'rocksdb::ObjectLibrary'
 1281 |         ObjectLibrary::PatternEntry("zenfs", false)
      |                        ^~~~~~~~~~~~
```

skyzh:
Should we also bump RocksDB version [here](https://github.com/westerndigitalcorporation/zenfs/blob/master/.github/workflows/smoke-test.yaml#L23)?

metaspace:
> Should we also bump RocksDB version [here](https://github.com/westerndigitalcorporation/zenfs/blob/master/.github/workflows/smoke-test.yaml#L23)?

Yes, that is the right place to bump RocksDB version. Right now we are held back by a bug in `db_stress`: https://github.com/westerndigitalcorporation/zenfs/issues/72. We need to fix that before bumping RocksDB version.

aravind-wdc:
Looks good!

yhr (changes requested):
This will break percona server and terarkdb integration as they are currently both locked to older versions of rocksdb. We need to figure out how to deal with the public API break.

SheldonZhong:
It seems fine for percona-server as long as rocksdb and zenfs are both locked to the compatible versions in their repo. They could just bump up rocksdb along with zenfs. 

As for terarkdb, the patch facebook/rocksdb#9264 might be able to applied on it without much effort. But we might need to customize an API for it if terarkdb starts to diverge from rocksdb in the future.

yhr:
The upstream API breakage has now been mended and zenfs registration works again. See https://github.com/facebook/rocksdb/commit/1973fcba111fc451df69eb2ec38eb66c285f316f

PR #128, opened 2022-01-03
fs: build failure when compiling rocksdb in debug mode

aravind-wdc:
When building with DEBUG_LEVEL=1 with below command,
"make clean;DEBUG_LEVEL=1 ROCKSDB_PLUGINS=zenfs make -j48 db_bench db_stress ldb"
rocksdb compilation fails with error:
"error: ‘std::string rocksdb::GetLogFilename(std::string)’ defined but not used"

Current logic defines the function GetLogFileName() for all debug levels(0,1,2)
but uses it only when debug-level=0. So for debug builds it gets defined but
is not used, resulting in the error. Also logs are enabled in production
builds now. Related to 7f8392e.

Define it such that is enabled for (debug-level > 0) OR when TERARKDB is defined.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -1144,6 +1144,7 @@ std::map<std::string, Env::WriteLifeTimeHint> ZenFS::GetWriteLifeTimeHints() {
   return hint_map;
 }
 
+#if !defined(NDEBUG) || defined(WITH_TERARKDB)
 static std::string GetLogFilename(std::string bdev) {
   std::ostringstream ss;
   time_t t = time(0);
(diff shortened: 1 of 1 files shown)


aravind-wdc:
Added the commits for fixing the apis as well to this one, with the review comments fom PR #93

yhr (changes requested):
@aravind-wdc : The debug-build fix looks good, but one PR per fix/feature, please! Drop the API fixes from this PR

aravind-wdc:
@yhr
I thought the patches are inter dependent, so I submitted in one PR.
Without fixing the debug build, we cannot compile/test the db_stress/db_bench issues (which need the api-fix, for db_stress), in debug mode.
So by creating 2 PR, it would create a dependency of one PR to be pulled before other. If you still think 2 PRs makes more sense, I can submit another PR. Pls let me know.

yhr (approved):

PR #130, opened 2022-01-05
fs: increase zone extent size to 64 bits

MaisenbacherD:
Increase the zone extent size to be 64 bits instead of 32, to
support very large file sizes in large zones.

This change is not backward-compatible with the current on-disk
format, so we need to step up the version.


MaisenbacherD:
The for_v2(https://github.com/westerndigitalcorporation/zenfs/tree/for_v2) branch can be deleted after merging this

damien-lemoal:
s/in stead/instead in the commit message

MaisenbacherD:
> s/in stead/instead in the commit message

Fixed :)

metaspace (approved):

yhr (approved):

PR #131, opened 2022-01-05
fs: Fix use of uninitialized value in `ZonedBlockDevice` destructor

metaspace:
Reported-by: Yura Sorokin <yura.sorokin@percona.com>

### DIFF
--- a/.github/workflows/cpp-lint.yaml
+++ b/.github/workflows/cpp-lint.yaml
@@ -2,9 +2,11 @@ on:
   push:
     branches:
     - master
+    - 1.0
   pull_request:
     branches:
     - master
+    - 1.0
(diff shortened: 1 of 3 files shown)


metaspace:
Going to enable smoke tests for 1.0

yhr (changes requested):
Looks good, but the commit message should say something about the constructor, not the destructor, right? :)

metaspace:
> Looks good, but the commit message should say something about the constructor, not the destructor, right? :)

Depends. The use that caused problems was on an error path through the destructor. The fix is in the constructor though.

yhr:
allright

yhr (approved):

Issue #110, opened 2021-12-09
Reduce size of the zenfs utility binary


yhr:
The size of the zenfs utility is ~200MB, we can strip that down to 50MB just by using strip.
Let's see if we can make it even smaller.

yhr:
fixed, #137 merged

PR #137, opened 2022-01-11
utils: strip debug symbols from zenfs utility binary

aravind-wdc:
zenfs utility is built with debug symbols, causing it to go more than
210MB in size. Modify the Makefile to compile it without the debug symbols
for normal build. This reduces the size from ~210MB to ~6MB.

Added a debug target so that "make debug", would still compile the binary
with the symbols. Now zenfs utility can be built in release mode by using
the command "make" or "make all", it can be built in debug mode using
"make debug".

Signed-off-by: Aravind Ramesh<Aravind.Ramesh@wdc.com>

### DIFF
--- a/util/Makefile
+++ b/util/Makefile
@@ -4,6 +4,7 @@ TARGET = zenfs
 
 CC ?= gcc
 CXX ?= g++
+OBJCOPY ?= objcopy
 
 CXXFLAGS = $(shell pkg-config --cflags rocksdb)
 LIBS = $(shell pkg-config --static --libs rocksdb)
@@ -11,10 +12,16 @@ LIBS = $(shell pkg-config --static --libs rocksdb)
 CXXFLAGS +=  $(EXTRA_CXXFLAGS)
 LDFLAGS +=  $(EXTRA_LDFLAGS)
 
-all: $(TARGET)
+all: $(TARGET) $(TARGET).dbg
+
+$(TARGET).dbg: $(TARGET)
+	@$(OBJCOPY) --only-keep-debug $(TARGET) $(TARGET).dbg
+	@$(OBJCOPY) --strip-all $(TARGET)
+
+debug: $(TARGET)
 
 $(TARGET): $(TARGET).cc
-	$(CXX) $(CXXFLAGS) -o $(TARGET) $< $(LIBS) $(LDFLAGS)
+	$(CXX) $(CXXFLAGS) -g -o $(TARGET) $< $(LIBS) $(LDFLAGS)
 
 clean:
-	$(RM) $(TARGET)
+	$(RM) $(TARGET) $(TARGET).dbg


aravind-wdc:
This fixes issue https://github.com/westerndigitalcorporation/zenfs/issues/110.

Without debug symbols(make):
```
rocksdb/plugin/zenfs/util$ ls -lh zenfs
-rwxrwxr-x 1 zns zns 6.3M Jan 11 23:58 zenfs

```
With Debug symbols (make debug):
```
rocksdb/plugin/zenfs/util$ ls -lh zenfs
-rwxrwxr-x 1 zns zns 216M Jan 11 23:59 zenfs

```

metaspace:
The `-O3` option is completely unrelated to removing debug symbols :shrug:  But fine to add it for release builds :)

It can be quite useful to have debug symbols for release builds. If we get a core dump or a backtrace from a user or something like that. We could make release builds with `-O3 -g` and then move the debug symbols to a separate file after build. Package maintainers can then choose whether to install debug symbols as a separate package, or not at all.

aravind-wdc:
> The `-O3` option is completely unrelated to removing debug symbols 🤷 But fine to add it for release builds :)

Yes.
Since the idea was to reduce the size(strip the binary of debug symbols), I thought we might as well optimize the binary for execution speed, and also generally release build binaries are optimized.

Using `-O3 -g` would still keep the binary size as 217MB. Are you saying we move the debug symbols to another file and then strip the binary, as a part of the regular build (make all) ?

metaspace:
> Using `-O3 -g` would still keep the binary size as 217MB. Are you saying we move the debug symbols to another file and then strip the binary, as a part of the regular build (make all) ?

Yes. It seems that there are may ways to do it: https://stackoverflow.com/questions/866721/how-to-generate-gcc-debug-symbol-outside-the-build-target

util/Makefile
```diff
@@ -4,17 +4,24 @@ TARGET = zenfs
...
+OBJCOPY ?= objcopy
 
 CXXFLAGS = $(shell pkg-config --cflags rocksdb)
 LIBS = $(shell pkg-config --static --libs rocksdb)
 
 CXXFLAGS +=  $(EXTRA_CXXFLAGS)
 LDFLAGS +=  $(EXTRA_LDFLAGS)
 
-all: $(TARGET)
+all: $(TARGET) generate_dbg
```
metaspace, line 15:
```suggestion
all: $(TARGET) $(TARGET).dbg
```

util/Makefile
```diff
@@ -4,17 +4,24 @@ TARGET = zenfs
...
+
+generate_dbg:
+	@$(OBJCOPY) --only-keep-debug $(TARGET) $(TARGET).dbg
+	@$(OBJCOPY) --strip-debug $(TARGET)
```
metaspace, line 19:
```suggestion
$(TARGET).dbg:
	@$(OBJCOPY) --only-keep-debug $(TARGET) $(TARGET).dbg
	@strip $(TARGET)
```

I think strip removes more stuff than objcpy --strip-debug. Could be that objcopy can do the same as strip with other options.

util/Makefile
```diff
@@ -4,17 +4,24 @@ TARGET = zenfs
...
+
+debug: $(TARGET)
```
metaspace, line 21:
```suggestion
debug: $(TARGET) CXXFLAGS+="-Og -g"
```
I don't know if the append works as a target specific variable, but I think it would be good to use these options for debug builds.

metaspace (changes requested):

metaspace:
I think sign-off is formatted wrong. You need a space between name and `<` to make the CI happy

util/Makefile
aravind-wdc: Removed the generate_dbg.

> I think sign-off is formatted wrong. You need a space between name and `<` to make the CI happy

Ok,  formatted correctly.

util/Makefile
aravind-wdc: I think using one command (objcopy or strip) in both places is better,  the strip man page also uses objcopy --strip-debug. Post strip, the difference in size is not huge between strip/objcopy.  I feel nothing wrong with using 
"objcopy --strip-debug"

util/Makefile
aravind-wdc: Added -g to main target, without it, gdb was not able to load symbols from the .dbg file.

util/Makefile
aravind-wdc: 'objcopy --strip-all' does the same thing as strip binary does, changed it.

yhr:
pushed manually

PR #138, opened 2022-01-12
fs: Fix ZoneFile:: GetIOType wrong return value

royguo:
fs: Fix ZoneFile:: GetIOType wrong return value


yhr (changes requested):
Looks good but could you change the subject to fs: Fix ZoneFile:: GetIOType wrong return value
and sign off the commit?

yhr (approved):

yhr:
Pushed manually

PR #142, opened 2022-01-18
Safer migration delete check

yhr:
Files may be deleted or renamed at any time we migrate files in
a garabage-collection thread. By checking IsDeleted rather
than checking if the file name exists, we make sure that
we we won't write spurious update records for a file
that is deleted and/or renamed while being migrated.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -441,9 +441,9 @@ IOStatus ZenFS::SyncFileMetadata(ZoneFile* zoneFile, bool replace) {
 
   std::lock_guard<std::mutex> lock(files_mtx_);
 
-  if (GetFileInternal(zoneFile->GetFilename()) == nullptr) {
-    Info(logger_, "File %s doesn't exist, skip sync file metadata!",
-         zoneFile->GetFilename().data());
+  if (zoneFile->IsDeleted()) {
+    Info(logger_, "File %s has been deleted, skip sync file metadata!",
+         zoneFile->GetFilename().c_str());
     return IOStatus::OK();
   }
 
@@ -1385,9 +1385,10 @@ IOStatus ZenFS::MigrateFileExtents(
   IOStatus s = IOStatus::OK();
   Info(logger_, "MigrateFileExtents, fname: %s, extent count: %lu",
        fname.data(), migrate_exts.size());
+
   // The file may be deleted by other threads, better double check.
   auto zfile = GetFile(fname);
-  if (zfile == nullptr || zfile->IsOpenForWR()) {
+  if (zfile == nullptr || zfile->IsDeleted() || zfile->IsOpenForWR()) {
     return IOStatus::OK();
   }
 
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -497,6 +497,8 @@ IOStatus ZenFS::DeleteFile(std::string fname) {
         /* Failed to persist the delete, return to a consistent state */
         files_.insert(std::make_pair(fname.c_str(), zoneFile));
       } else {
+        /* Mark up the file as deleted so it won't be migrated by GC */
+        zoneFile->SetDeleted();
         zoneFile.reset();
       }
     } else {
--- a/fs/io_zenfs.h
+++ b/fs/io_zenfs.h
@@ -70,6 +70,7 @@ class ZoneFile {
   bool open_for_wr_ = false;
   time_t m_time_;
   bool is_sparse_ = false;
+  bool is_deleted_ = false;
 
   MetadataWriter* metadata_writer_ = NULL;
 
@@ -145,8 +146,10 @@ class ZoneFile {
   IOStatus CloseActiveZone();
 
  public:
-  std::shared_ptr<ZenFSMetrics> GetZBDMetrics() { return zbd_->GetMetrics(); }
-  IOType GetIOType() const { return io_type_; }
+  std::shared_ptr<ZenFSMetrics> GetZBDMetrics() { return zbd_->GetMetrics(); };
+  IOType GetIOType() const { return io_type_; };
+  bool IsDeleted() const { return is_deleted_; };
+  void SetDeleted() { is_deleted_ = true; };
   IOStatus RecoverSparseExtents(uint64_t start, uint64_t end, Zone* zone);
 
  public:


MaisenbacherD (commented):
In ZenFS::DeleteFiles the zoneFile is checked if it still exists: [ if (zoneFile != nullptr) {](https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L488).
This check should be extended to `if (zoneFile != nullptr && zoneFile->IsDeleted())`

MaisenbacherD:
This line might cause issues when the IsDeleted flag is set: https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L526

MaisenbacherD:
Checks if a file exists should be extended to check if `zoneFile->IsDeleted()`
https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L572
https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L592
https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L662
https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L768
https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L859

yhr:
> This line might cause issues when the IsDeleted flag is set: https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L526
> 
> Same here: https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L544

Why?

yhr:
> In ZenFS::DeleteFiles the zoneFile is checked if it still exists: [ if (zoneFile != nullptr) {](https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L488). This check should be extended to `if (zoneFile != nullptr && zoneFile->IsDeleted())`

Are you worried about a race? we hold the file mutex in DeleteFile when we call GetFileInternal, so why do we need to check the delte flag?

MaisenbacherD:
> > In ZenFS::DeleteFiles the zoneFile is checked if it still exists: [ if (zoneFile != nullptr) {](https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L488). This check should be extended to `if (zoneFile != nullptr && zoneFile->IsDeleted())`
> 
> Are you worried about a race? we hold the file mutex in DeleteFile when we call GetFileInternal, so why do we need to check the delte flag?

You wrote in the commit message: 
> ...
As the zone files are shared pointers, readers
and migration threads might still hold a reference
to the file even though the file is deleted. ...

By extending the check we might avoid a similar pattern in the future. 

I just realised that I made a mistake in my proposal. The check should be `if (zoneFile != nullptr && !zoneFile->IsDeleted())`

MaisenbacherD:
> > This line might cause issues when the IsDeleted flag is set: https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L526
> > Same here: https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/fs_zenfs.cc#L544
> 
> Why?

Same reason as for the commit message:
> ...
As the zone files are shared pointers, readers
and migration threads might still hold a reference
to the file even though the file is deleted. ...

yhr:
@MaisenbacherD : It's fine if readers who have opened a file can continue to access it, even if the file has been deleted by another process. This is expected behavior. This PR should not change that.

fs/fs_zenfs.cc
```diff
@@ -441,9 +441,9 @@ IOStatus ZenFS::SyncFileMetadata(ZoneFile* zoneFile, bool replace) {
 
   std::lock_guard<std::mutex> lock(files_mtx_);
 
-  if (GetFileInternal(zoneFile->GetFilename()) == nullptr) {
-    Info(logger_, "File %s doesn't exist, skip sync file metadata!",
-         zoneFile->GetFilename().data());
+  if (zoneFile->IsDeleted()) {
```
royguo, line 444: If we are holding the files_mtx_, and the `GetFileInternal` returns true, then this file surely exists. But if we simply check `is_deleted_` flag here, it may return false (means the file exists), but the file actually had been deleted by other threads, in this case, we may migrate a deleted file.

royguo: Oh, never mind, I just realized the `DeleteFile` is also holding file_mtx_

MaisenbacherD:
> @MaisenbacherD : It's fine if readers who have opened a file can continue to access it, even if the file has been deleted by another process. This is expected behavior. This PR should not change that.

My comments are resolved. As you indicated `GetFile` locks the `files_mtx_`, thus if a file was deleted it is removed from `files_` at that point.

fs/fs_zenfs.cc
```diff
@@ -1385,9 +1385,10 @@ IOStatus ZenFS::MigrateFileExtents(
   IOStatus s = IOStatus::OK();
   Info(logger_, "MigrateFileExtents, fname: %s, extent count: %lu",
        fname.data(), migrate_exts.size());
+
   // The file may be deleted by other threads, better double check.
   auto zfile = GetFile(fname);
```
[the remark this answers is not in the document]
yhr: @MaisenbacherD : What is the potential race? if we hold a reference to the file, it existed at that point and because we have a reference to it will continue to be valid, so it should safe to check if it was deleted after that right?

fs/fs_zenfs.cc
```diff
@@ -1383,9 +1385,10 @@ IOStatus ZenFS::MigrateFileExtents(
...
-  if (zfile == nullptr || zfile->IsOpenForWR()) {
+  if (zfile == nullptr || zfile->IsDeleted() || zfile->IsOpenForWR()) {
```
metaspace, line 1391: What is the purpose of `zfile->IsDeleted()` here? If we are holding the `files_mtx_`, `GetFile(fname)` will return `nullptr` if the file is deleted, and `IsDeleted()` has no effect.

If we are not holding the mutex, we will potentially observe out of order behavior of `GetFile()` and `IsDeleted()`. so file will appear in the files list and is deleted may return false for a file that is in fact deleted.

metaspace: So the purpose seems to be to guard against the file being deleted or renamed after the file is selected for migration, but before migration is effected. While we do hold the mutex protecting the file map while we query the map for the file name, the lock is released when `GetFile()` returns. If the file was renamed or deleted at this point, `GetFile()` would return a `nullptr`, so no need to check `IsDeleted()`. If the file is deleted or renamed after `GetFile()` returns, this code is not guaranteed to observe the change of the `is_deleted_` field, and we may continue working with a file that is deleted.

If it is a problem that a file is deleted or renamed while migration is ongoing (which I assume, otherwise the checks are not needed), we need to synchronize on those operations, blocking until migration is done.

metaspace (changes requested):

yhr:
Looking at your review comments and going through the code again, I've convinced myself that this change is not needed - rename, deletes and file metadata syncs all hold the files mutex, so any renames/deletes would be in sync across threads.

PR #143, opened 2022-01-21
metrics: add more metrics explanition comments & refine metrics

royguo:
- The previous metrics implementation is not well documented
- Some metrics helper functions are not necessary (e.g. `ZENFS_LABEL` confuses readers)
- Add QPS and throughput trace for ZoneFile::PositionedAppend
- Add Throughput trace for Zone::Append

### DIFF
--- a/fs/metrics_sample.h
+++ b/fs/metrics_sample.h
@@ -1,4 +1,4 @@
-//  SPDX-License-Identifier: Apache License 2.0 OR GPL-2.0
+// SPDX-License-Identifier: Apache License 2.0 OR GPL-2.0
 
 #include <iostream>
 #include <unordered_map>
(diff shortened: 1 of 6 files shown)


fs/metrics_sample.h
```diff
@@ -1,4 +1,9 @@
-//  SPDX-License-Identifier: Apache License 2.0 OR GPL-2.0
```
yhr, line 1: Why change the spdx license identifier to a full license text? It has the same meaning. We want to use the spdx-licenses  for easier license classification.

yhr (changes requested):
One small comment

fs/metrics_sample.h
royguo: I formated the comments with my default IDE setting....will change it back

yhr (approved):

PR #145, opened 2022-01-28
fs: ensure that metadata is persisted before releasing active zone

yhr:
If the active zone is released and the file has not been marked
up as closed (active extent = NO_EXTENT) the file recovery
process will try to recover everything up to the active extents
zone write pointer, which might contain data from other files.

To make sure this does not happen, mark up the file as closed
before releasing the active zone.


metaspace (approved):
Looks good to me. Probably an very small chance to trigger this bug, but good catch!

PR #147, opened 2022-01-28
fs: Align direct and unaligned read requests to blocksize boundary

aravind-wdc:
For read requests which are direct and unaligned, zenfs fallsback
on buffered reads. As a result, applications using direct reads
could experience buffered reads and can also see a spike in
kernel buffer usage.
Align the read request size to the next blocksize boundary and
issue direct reads for such requests.

### DIFF
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -355,9 +355,17 @@ IOStatus ZoneFile::PositionedRead(uint64_t offset, size_t n, Slice* result,
     if ((pread_sz + r_off) > extent_end) pread_sz = extent_end - r_off;
 
     /* We may get some unaligned direct reads due to non-aligned extent lengths,
-     * so fall back on non-direct-io in that case.
+     * so increase read request size to be aligned to next blocksize boundary.
      */
     bool aligned = (pread_sz % zbd_->GetBlockSize() == 0);
+
+    size_t bytes_to_align = 0;
+    if (direct && !aligned) {
+      bytes_to_align = zbd_->GetBlockSize() - (pread_sz % zbd_->GetBlockSize());
+      pread_sz += bytes_to_align;
+      aligned = true;
+    }
+
     if (direct && aligned) {
       r = pread(f_direct, ptr, pread_sz, r_off);
     } else {
@@ -371,7 +379,13 @@ IOStatus ZoneFile::PositionedRead(uint64_t offset, size_t n, Slice* result,
       break;
     }
 
-    pread_sz = (size_t)r;
+    /* Verify and update the the bytes read count (if read size was incremented,
+     * for alignment purposes).
+     */
+    if ((size_t)r <= pread_sz - bytes_to_align)
+      pread_sz = (size_t)r;
+    else
+      pread_sz -= bytes_to_align;
 
     ptr += pread_sz;
     read += pread_sz;


aravind-wdc:
This address the issue WDCZEN-22 in jira, where percona-myrocks experienced high kernel memory usage (~100gb) in bulk mode when doing direct reads.

fs/io_zenfs.cc
```diff
@@ -355,9 +355,17 @@ IOStatus ZoneFile::PositionedRead(uint64_t offset, size_t n, Slice* result,
...
-     * so fall back on non-direct-io in that case.
+     * so increase read request size to be aligned to next blocksize boundary.
      */
     bool aligned = (pread_sz % zbd_->GetBlockSize() == 0);
+
+    size_t bytes_to_align = 0;
+    if (direct && !aligned) {
+      bytes_to_align = zbd_->GetBlockSize() - (pread_sz % zbd_->GetBlockSize());
+      pread_sz += bytes_to_align;
+      aligned = (pread_sz % zbd_->GetBlockSize() == 0);
```
yhr, line 366: Why not just set aligned to true?

fs/io_zenfs.cc
```diff
@@ -371,7 +379,13 @@ IOStatus ZoneFile::PositionedRead(uint64_t offset, size_t n, Slice* result,
...
     }
 
-    pread_sz = (size_t)r;
+    /* Verify and update the the bytes read count (if read size was incremented,
+     * for alignment purposes).
+     */
+    if ((size_t)r <= pread_sz - bytes_to_align)
+      pread_sz = (size_t)r;
+    else
+      pread_sz -= bytes_to_align;
```
yhr, line 388: This won't work if only part of the alignment area has been read. This *should* not happen for direct reads but let's make sure we don't fail to calculate the right result in that case anyway.

Keep pread_sz indicating the valid data bytes and introduce a pread_sz_aligned varible in stead and use that for the direct read. then it is enough to do this:

if (r < pread_sz  ) pread_sz = r;
(when r >= pread_sz we don't need to do anything)

yhr (changes requested):

fs/io_zenfs.cc
aravind-wdc: Ok. Will update this.

fs/io_zenfs.cc
aravind-wdc: Just to understand the case we are discussing here, if the initial pread_sz is 6K, then we align pread_sz to 8K (bytes_to_align=2K), then pread() returns 7K as the bytes read.
I considered this case where only part of alignment area is read and here I return, the bytes requested initially, (i.e. 6K (8K - 2k) in this example), so that it is like we returned exactly what was initially requested and whatever extra bytes were read due to alignment adjustment is just ignored. 

Even with 
r = pread(f_direct, ptr, pread_sz_aligned, r_off);
if (r < pread_sz ) pread_sz = r;

If r was 7K, then pread_sz will be 6K. So in both cases, 6K  will be the bytes read(for accounting purposes).
So I think the current implementation behaves the same way as suggested changes.

Or, are you suggesting we do;
if (r < pread_sz_aligned) pread_sz = r;
To return 7K when 6K was the initial request.

yhr: Ah, yes, you're right.

yhr (approved):
Looks good to me!

yhr:
@aravind-wdc : could you submit a PR to the V1 branch as well?

yhr:
pushed manually

PR #149, opened 2022-02-01
Make zenfs compatible with latest plugin registration API

yhr:
Upstream RocksDB has changed the filesystem registration API.
Add support for the new API and enable compatibility with
older versions of RocksDB (version < 7 ) by passing a compile time
flag (ZENFS_USE_OLD_REGISTRATION_API).

Example(building zenfs with rocksdb 6.x.x after this commit)

CFLAGS=-DZENFS_USE_OLD_REGISTRATION_API ROCKSDB_PLUGINS="zenfs" make db_bench

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -995,8 +995,6 @@ Status ZenFS::RecoverFrom(ZenMetaLog* log) {
     return Status::NotFound("ZenFS", "No snapshot found");
 }
 
-#define ZENV_URI_PATTERN "zenfs://"
-
 /* Mount the filesystem by recovering form the latest valid metadata zone */
 Status ZenFS::Mount(bool readonly) {
   std::vector<Zone*> metazones = zbd_->GetMetaZones();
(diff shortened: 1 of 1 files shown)


yhr:
CI needs to set CFLAGS=-DZENFS_USE_OLD_REGISTRATION_API when building for the current version we're locked to

fs/fs_zenfs.cc
```diff
@@ -1459,9 +1457,17 @@ IOStatus ZenFS::MigrateFileExtents(
 extern "C" FactoryFunc<FileSystem> zenfs_filesystem_reg;
 
 FactoryFunc<FileSystem> zenfs_filesystem_reg =
+#ifdef ZENFS_USE_OLD_REGISTRATION_API
```
metaspace, line 1460:
```suggestion
if (ROCKSDB_MAJOR < 7) {...
```
I think it would be more user friendly to use the defined version macros. Using conditional c expression instead of conditional compile should be ok, the compiler will strip the dead code.

metaspace (changes requested):

fs/fs_zenfs.cc
yhr: Uhm, how is this supposed to work? How could we conditionally define  the registration in runtime?(!)

yhr: But hey, i just learned that we CAN get the preprocessor to do it.
```


#if ROCKSDB_MAJOR < 7
..
# else
..
#endif
```

yhr: problem is that we're not at 7 yet, maybe we can compare with 6.29.0

metaspace: > If you get that to compile.. :)

Well, add the rest of the conditional brackets, etc. why would it not compile?

metaspace: > problem is that we're not at 7 yet, maybe we can compare with 6.29.0

Problem with that approach is that FB may release a version 6.29.0 with the old API, and then we break. They should probably have updated the major when they did the breaking change :shrug: We could raise an issue in rocksdb tracker to ask about that.

yhr: @metaspace : I just pushed an update, i think the latest version is safe, i checked pre/post breaking change and it seems ok

metaspace (approved):

PR #150, opened 2022-02-03
fs: Fix a bug where `ZonedWritableFile` bufferes are not flushed

metaspace:
If a `ZonedWritableFile` is destroyed without being closed, buffered contents
are not flushed to disk. This patch activates close procedure on object
destruction to fix the problem.

### DIFF
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -704,7 +704,7 @@ ZonedWritableFile::ZonedWritableFile(ZonedBlockDevice* zbd, bool _buffered,
 }
 
 ZonedWritableFile::~ZonedWritableFile() {
-  IOStatus s = zoneFile_->CloseWR();
+  IOStatus s = CloseInternal();
   if (buffered) free(sparse_buffer);
 
   if (!s.ok()) {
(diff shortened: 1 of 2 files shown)


metaspace:
> Looks good to me, how did you discover the issue?

I was reviewing an unrelated PR. Not sure if we are able to trigger this one at the moment (did not check) but we might in the future.

fs/io_zenfs.cc
```diff
@@ -699,7 +699,7 @@ ZonedWritableFile::ZonedWritableFile(ZonedBlockDevice* zbd, bool _buffered,
 }
 
 ZonedWritableFile::~ZonedWritableFile() {
-  IOStatus s = zoneFile_->CloseWR();
+  IOStatus s = CloseInternal();
```
yhr, line 702: This will trigger an unneeded metadata write in the normal case, can you make this conditional on the zone file being open for write?

yhr (changes requested):
Pardon, had another look over the change while rebasing and found one issue.

fs/io_zenfs.cc
yhr: I meant that don't call closeinternal(generating another metadata sync) unless the file is still open

metaspace: Since we don't want metadata sync on the normal close path if the file is already closed, I put the check in `CloseInternal()`. I check on `zoneFile_->IsOpenForWR()` since there is no tracking in `ZonedWritableFile`.

yhr (approved):

PR #153, opened 2022-02-07
Integration fixes for percona-server

yhr:


MaisenbacherD (commented):
I have one comment, otherwise it looks good to me :)

fs/fs_zenfs.cc
```diff
@@ -1383,9 +1383,10 @@ IOStatus ZenFS::MigrateFileExtents(
   IOStatus s = IOStatus::OK();
   Info(logger_, "MigrateFileExtents, fname: %s, extent count: %lu",
        fname.data(), migrate_exts.size());
+
   // The file may be deleted by other threads, better double check.
   auto zfile = GetFile(fname);
-  if (zfile == nullptr || zfile->IsOpenForWR()) {
+  if (zfile == nullptr || zfile->IsDeleted() || zfile->IsOpenForWR()) {
```
metaspace, line 1388: Maybe add a comment that the `IsDeleted` check is an optimistic optimization. Code is also valid without the check, but the check can help avoid redundant work.

metaspace (approved):
Looks good. Only have the one comment about the purpose of the `IsDeleted()` check when migrating extents.

fs/io_zenfs.cc
```diff
@@ -446,6 +446,60 @@ IOStatus ZoneFile::AllocateNewZone() {
   return PersistMetadata();
 }
 
+/* Byte-aligned writes without a sparse header */
+IOStatus ZoneFile::BufferedAppend(char* buffer, uint32_t data_size) {
```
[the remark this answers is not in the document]
yhr: Yes, I'm planning to do that as a follow up

fs/fs_zenfs.cc
yhr: Will do, i'll rework the commit message as well to clarify the isdeleted usage

metaspace: I think this was dropped, so resolving comment.

aravind-wdc:
Looks good to me.

PR #154, opened 2022-02-08
fs: drop empty-file assertion for writable files

yhr:
We do allow that files can be re-opened for appending
(ReopenWritableFile), so drop the assertion that the
file is empty when creating a new writable file.


MaisenbacherD:
Looks good to me!

yhr:
I checked, and the reason for us having to change the implementation in the first place (allowing appends) was an API redefinition (!) in rocksdb last october, see https://github.com/facebook/rocksdb/commit/a282eff3d14b6881485b188a2a07b8a39e452898

PR #156, opened 2022-02-16
fs: Add LinkFile api support to ZenFS.

aravind-wdc:
This api simulates posix hard link behavior.
It accepts the source-file and destination-file names, if the source file is
available, it creates an entry in the in-memory files_ map and also
updates the source-file's metadata about the newly linked file and
increments the link count for the source-file and persists this information
on disk.
File deletion ensures that as long as there are links associated to a file,
the file is not deleted on disk and only link specific metadata is updated
and persisted on disk.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -499,14 +499,17 @@ IOStatus ZenFS::DeleteFileNoLock(std::string fname, const IOOptions& options,
     if (zoneFile->IsOpenForWR())
       return IOStatus::Busy("ZenFS::DeleteFileNoLock(): file open for writing:",
                             fname.c_str());
-    zoneFile = files_[fname];
     files_.erase(fname);
-    EncodeFileDeletionTo(zoneFile, &record);
+    s = zoneFile->RemoveLinkName(fname);
+    if (!s.ok()) return s;
+    EncodeFileDeletionTo(zoneFile, &record, fname);
     s = PersistRecord(record);
     if (!s.ok()) {
       /* Failed to persist the delete, return to a consistent state */
       files_.insert(std::make_pair(fname.c_str(), zoneFile));
+      zoneFile->AddLinkName(fname);
     } else {
+      if (zoneFile->GetNrLinks() > 0) return s;
       /* Mark up the file as deleted so it won't be migrated by GC */
       zoneFile->SetDeleted();
       zoneFile.reset();
@@ -693,8 +696,9 @@ IOStatus ZenFS::OpenWritableFile(const std::string& fname,
       resetIOZones = true;
     }
 
-    zoneFile = std::make_shared<ZoneFile>(zbd_, fname, next_file_id_++);
+    zoneFile = std::make_shared<ZoneFile>(zbd_, next_file_id_++);
     zoneFile->SetFileModificationTime(time(0));
+    zoneFile->AddLinkName(fname);
 
     /* RocksDB does not set the right io type(!)*/
     if (ends_with(fname, ".log")) {
@@ -793,15 +797,18 @@ IOStatus ZenFS::RenameFileNoLock(const std::string& source_path,
       }
     }
 
+    s = source_file->RenameLink(source_path, dest_path);
+    if (!s.ok()) return s;
     files_.erase(source_path);
-    source_file->Rename(dest_path);
+
     files_.insert(std::make_pair(dest_path, source_file));
 
     s = SyncFileMetadataNoLock(source_file);
     if (!s.ok()) {
       /* Failed to persist the rename, roll back */
       files_.erase(dest_path);
-      source_file->Rename(source_path);
+      s = source_file->RenameLink(dest_path, source_path);
+      if (!s.ok()) return s;
       files_.insert(std::make_pair(source_path, source_file));
     }
   } else {
@@ -821,6 +828,35 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
     s = RenameFileNoLock(source_path, dest_path, options, dbg);
   }
   if (s.ok()) s = zbd_->ResetUnusedIOZones();
+  return s;
+}
+
+IOStatus ZenFS::LinkFile(const std::string& fname, const std::string& lname,
+                         const IOOptions& options, IODebugContext* dbg) {
+  std::shared_ptr<ZoneFile> src_file(nullptr);
+  IOStatus s;
+
+  Debug(logger_, "LinkFile: %s to %s\n", fname.c_str(), lname.c_str());
+  {
+    std::lock_guard<std::mutex> lock(files_mtx_);
+
+    if (GetFileNoLock(lname) != nullptr)
+      return IOStatus::InvalidArgument("Failed to create link, target exists");
+
+    src_file = GetFileNoLock(fname);
+    if (src_file != nullptr) {
+      src_file->AddLinkName(lname);
+      files_.insert(std::make_pair(lname, src_file));
+      s = SyncFileMetadataNoLock(src_file);
+      if (!s.ok()) {
+        s = src_file->RemoveLinkName(lname);
+        if (!s.ok()) return s;
+        files_.erase(lname);
+      }
+      return s;
+    }
+  }
+  s = target()->LinkFile(ToAuxPath(fname), ToAuxPath(lname), options, dbg);
   return s;
 }
 
@@ -853,7 +889,7 @@ void ZenFS::EncodeJson(std::ostream& json_stream) {
 }
 
 Status ZenFS::DecodeFileUpdateFrom(Slice* slice, bool replace) {
-  std::shared_ptr<ZoneFile> update(new ZoneFile(zbd_, "not_set", 0));
+  std::shared_ptr<ZoneFile> update(new ZoneFile(zbd_, 0));
   uint64_t id;
   Status s;
 
@@ -867,17 +903,20 @@ Status ZenFS::DecodeFileUpdateFrom(Slice* slice, bool replace) {
   for (auto it = files_.begin(); it != files_.end(); it++) {
     std::shared_ptr<ZoneFile> zFile = it->second;
     if (id == zFile->GetID()) {
-      std::string oldName = zFile->GetFilename();
+      for (const auto& name : zFile->GetLinkFiles()) {
+        if (files_.find(name) != files_.end())
+          files_.erase(name);
+        else
+          return Status::Corruption("DecodeFileUpdateFrom: missing link file");
+      }
 
       s = zFile->MergeUpdate(update, replace);
       update.reset();
 
       if (!s.ok()) return s;
 
-      if (zFile->GetFilename() != oldName) {
-        files_.erase(oldName);
-        files_.insert(std::make_pair(zFile->GetFilename(), zFile));
-      }
+      for (const auto& name : zFile->GetLinkFiles())
+        files_.insert(std::make_pair(name, zFile));
 
       return Status::OK();
     }
@@ -896,24 +935,26 @@ Status ZenFS::DecodeSnapshotFrom(Slice* input) {
   assert(files_.size() == 0);
 
   while (GetLengthPrefixedSlice(input, &slice)) {
-    std::shared_ptr<ZoneFile> zoneFile(new ZoneFile(zbd_, "not_set", 0));
+    std::shared_ptr<ZoneFile> zoneFile(new ZoneFile(zbd_, 0));
     Status s = zoneFile->DecodeFrom(&slice);
     if (!s.ok()) return s;
 
-    files_.insert(std::make_pair(zoneFile->GetFilename(), zoneFile));
     if (zoneFile->GetID() >= next_file_id_)
       next_file_id_ = zoneFile->GetID() + 1;
+
+    for (const auto& name : zoneFile->GetLinkFiles())
+      files_.insert(std::make_pair(name, zoneFile));
   }
 
   return Status::OK();
 }
 
 void ZenFS::EncodeFileDeletionTo(std::shared_ptr<ZoneFile> zoneFile,
-                                 std::string* output) {
+                                 std::string* output, std::string linkf) {
   std::string file_string;
 
   PutFixed64(&file_string, zoneFile->GetID());
-  PutLengthPrefixedSlice(&file_string, Slice(zoneFile->GetFilename()));
+  PutLengthPrefixedSlice(&file_string, Slice(linkf));
 
   PutFixed32(output, kFileDeletion);
   PutLengthPrefixedSlice(output, Slice(file_string));
@@ -923,6 +964,7 @@ Status ZenFS::DecodeFileDeletionFrom(Slice* input) {
   uint64_t fileID;
   std::string fileName;
   Slice slice;
+  IOStatus s;
 
   if (!GetFixed64(input, &fileID))
     return Status::Corruption("Zone file deletion: file id missing");
@@ -939,7 +981,9 @@ Status ZenFS::DecodeFileDeletionFrom(Slice* input) {
     return Status::Corruption("Zone file deletion: file ID missmatch");
 
   files_.erase(fileName);
-  zoneFile.reset();
+  s = zoneFile->RemoveLinkName(fileName);
+  if (!s.ok())
+    return Status::Corruption("Zone file deletion: file links missmatch");
 
   return Status::OK();
 }
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -860,6 +860,25 @@ IOStatus ZenFS::LinkFile(const std::string& fname, const std::string& lname,
   return s;
 }
 
+IOStatus ZenFS::NumFileLinks(const std::string& fname, const IOOptions& options,
+                             uint64_t* nr_links, IODebugContext* dbg) {
+  std::shared_ptr<ZoneFile> src_file(nullptr);
+  IOStatus s;
+
+  Debug(logger_, "NumFileLinks: %s\n", fname.c_str());
+  {
+    std::lock_guard<std::mutex> lock(files_mtx_);
+
+    src_file = GetFileNoLock(fname);
+    if (src_file != nullptr) {
+      *nr_links = (uint64_t)src_file->GetNrLinks();
+      return IOStatus::OK();
+    }
+  }
+  s = target()->NumFileLinks(ToAuxPath(fname), options, nr_links, dbg);
+  return s;
+}
+
 void ZenFS::EncodeSnapshotTo(std::string* output) {
   std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   std::string files_string;
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -879,6 +879,31 @@ IOStatus ZenFS::NumFileLinks(const std::string& fname, const IOOptions& options,
   return s;
 }
 
+IOStatus ZenFS::AreFilesSame(const std::string& fname, const std::string& link,
+                             const IOOptions& options, bool* res,
+                             IODebugContext* dbg) {
+  std::shared_ptr<ZoneFile> src_file(nullptr);
+  std::shared_ptr<ZoneFile> dst_file(nullptr);
+  IOStatus s;
+
+  Debug(logger_, "AreFilesSame: %s, %s\n", fname.c_str(), link.c_str());
+
+  {
+    std::lock_guard<std::mutex> lock(files_mtx_);
+    src_file = GetFileNoLock(fname);
+    dst_file = GetFileNoLock(link);
+    if (src_file != nullptr && dst_file != nullptr) {
+      if (src_file->GetID() == dst_file->GetID())
+        *res = true;
+      else
+        *res = false;
+      return IOStatus::OK();
+    }
+  }
+  s = target()->AreFilesSame(fname, link, options, res, dbg);
+  return s;
+}
+
 void ZenFS::EncodeSnapshotTo(std::string* output) {
   std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   std::string files_string;
--- a/fs/fs_zenfs.h
+++ b/fs/fs_zenfs.h
@@ -176,7 +176,7 @@ class ZenFS : public FileSystemWrapper {
 
   void EncodeSnapshotTo(std::string* output);
   void EncodeFileDeletionTo(std::shared_ptr<ZoneFile> zoneFile,
-                            std::string* output);
+                            std::string* output, std::string linkf);
 
   Status DecodeSnapshotFrom(Slice* input);
   Status DecodeFileUpdateFrom(Slice* slice, bool replace = false);
@@ -266,6 +266,10 @@ class ZenFS : public FileSystemWrapper {
   virtual IOStatus DeleteFile(const std::string& fname,
                               const IOOptions& options,
                               IODebugContext* dbg) override;
+  virtual IOStatus LinkFile(const std::string& fname, const std::string& lname,
+                            const IOOptions& options,
+                            IODebugContext* dbg) override;
+
   IOStatus GetFileSize(const std::string& f, const IOOptions& options,
                        uint64_t* size, IODebugContext* dbg) override;
   IOStatus RenameFile(const std::string& f, const std::string& t,
@@ -380,13 +384,6 @@ class ZenFS : public FileSystemWrapper {
       std::unique_ptr<MemoryMappedFileBuffer>* /*result*/) override {
     return IOStatus::NotSupported(
         "MemoryMappedFileBuffer is not implemented in ZenFS");
-  }
-
-  virtual IOStatus LinkFile(const std::string& /*src*/,
-                            const std::string& /*target*/,
-                            const IOOptions& /*options*/,
-                            IODebugContext* /*dbg*/) override {
-    return IOStatus::NotSupported("LinkFile is not supported in ZenFS");
   }
 
   virtual IOStatus NumFileLinks(const std::string& /*fname*/,
--- a/fs/fs_zenfs.h
+++ b/fs/fs_zenfs.h
@@ -269,6 +269,9 @@ class ZenFS : public FileSystemWrapper {
   virtual IOStatus LinkFile(const std::string& fname, const std::string& lname,
                             const IOOptions& options,
                             IODebugContext* dbg) override;
+  virtual IOStatus NumFileLinks(const std::string& fname,
+                                const IOOptions& options, uint64_t* nr_links,
+                                IODebugContext* dbg) override;
 
   IOStatus GetFileSize(const std::string& f, const IOOptions& options,
                        uint64_t* size, IODebugContext* dbg) override;
@@ -384,14 +387,6 @@ class ZenFS : public FileSystemWrapper {
       std::unique_ptr<MemoryMappedFileBuffer>* /*result*/) override {
     return IOStatus::NotSupported(
         "MemoryMappedFileBuffer is not implemented in ZenFS");
-  }
-
-  virtual IOStatus NumFileLinks(const std::string& /*fname*/,
-                                const IOOptions& /*options*/,
-                                uint64_t* /*count*/,
-                                IODebugContext* /*dbg*/) override {
-    return IOStatus::NotSupported(
-        "Getting number of file links is not supported in ZenFS");
   }
 
   virtual IOStatus AreFilesSame(const std::string& /*first*/,
--- a/fs/fs_zenfs.h
+++ b/fs/fs_zenfs.h
@@ -272,6 +272,10 @@ class ZenFS : public FileSystemWrapper {
   virtual IOStatus NumFileLinks(const std::string& fname,
                                 const IOOptions& options, uint64_t* nr_links,
                                 IODebugContext* dbg) override;
+  virtual IOStatus AreFilesSame(const std::string& fname,
+                                const std::string& link,
+                                const IOOptions& options, bool* res,
+                                IODebugContext* dbg) override;
 
   IOStatus GetFileSize(const std::string& f, const IOOptions& options,
                        uint64_t* size, IODebugContext* dbg) override;
@@ -387,13 +391,6 @@ class ZenFS : public FileSystemWrapper {
       std::unique_ptr<MemoryMappedFileBuffer>* /*result*/) override {
     return IOStatus::NotSupported(
         "MemoryMappedFileBuffer is not implemented in ZenFS");
-  }
-
-  virtual IOStatus AreFilesSame(const std::string& /*first*/,
-                                const std::string& /*second*/,
-                                const IOOptions& /*options*/, bool* /*res*/,
-                                IODebugContext* /*dbg*/) override {
-    return IOStatus::NotSupported("AreFilesSame is not supported in ZenFS");
   }
 
   void GetZenFSSnapshot(ZenFSSnapshot& snapshot,
--- a/fs/io_zenfs.cc
+++ b/fs/io_zenfs.cc
@@ -55,21 +55,19 @@ void ZoneExtent::EncodeJson(std::ostream& json_stream) {
 
 enum ZoneFileTag : uint32_t {
   kFileID = 1,
-  kFileName = 2,
+  kFileNameDeprecated = 2,
   kFileSize = 3,
   kWriteLifeTimeHint = 4,
   kExtent = 5,
   kModificationTime = 6,
   kActiveExtentStart = 7,
   kIsSparse = 8,
+  kLinkedFilename = 9,
 };
 
 void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
   PutFixed32(output, kFileID);
   PutFixed64(output, file_id_);
-
-  PutFixed32(output, kFileName);
-  PutLengthPrefixedSlice(output, Slice(filename_));
 
   PutFixed32(output, kFileSize);
   PutFixed64(output, fileSize);
@@ -99,15 +97,22 @@ void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
   if (is_sparse_) {
     PutFixed32(output, kIsSparse);
   }
+
+  for (uint32_t i = 0; i < linkfiles_.size(); i++) {
+    PutFixed32(output, kLinkedFilename);
+    PutLengthPrefixedSlice(output, Slice(linkfiles_[i]));
+  }
 }
 
 void ZoneFile::EncodeJson(std::ostream& json_stream) {
   json_stream << "{";
   json_stream << "\"id\":" << file_id_ << ",";
-  json_stream << "\"filename\":\"" << filename_ << "\",";
   json_stream << "\"size\":" << fileSize << ",";
   json_stream << "\"hint\":" << lifetime_ << ",";
   json_stream << "\"extents\":[";
+
+  for (const auto& name : GetLinkFiles())
+    json_stream << "\"filename\":\"" << name << "\",";
 
   bool first_element = true;
   for (ZoneExtent* extent : extents_) {
@@ -136,13 +141,6 @@ Status ZoneFile::DecodeFrom(Slice* input) {
     if (!GetFixed32(input, &tag)) break;
 
     switch (tag) {
-      case kFileName:
-        if (!GetLengthPrefixedSlice(input, &slice))
-          return Status::Corruption("ZoneFile", "Filename missing");
-        filename_ = slice.ToString();
-        if (filename_.length() == 0)
-          return Status::Corruption("ZoneFile", "Zero length filename");
-        break;
       case kFileSize:
         if (!GetFixed64(input, &fileSize))
           return Status::Corruption("ZoneFile", "Missing file size");
@@ -182,6 +180,15 @@ Status ZoneFile::DecodeFrom(Slice* input) {
       case kIsSparse:
         is_sparse_ = true;
         break;
+      case kLinkedFilename:
+        if (!GetLengthPrefixedSlice(input, &slice))
+          return Status::Corruption("ZoneFile", "LinkFilename missing");
+
+        if (slice.ToString().length() == 0)
+          return Status::Corruption("ZoneFile", "Zero length Linkfilename");
+
+        linkfiles_.push_back(slice.ToString());
+        break;
       default:
         return Status::Corruption("ZoneFile", "Unexpected tag");
     }
@@ -195,7 +202,6 @@ Status ZoneFile::MergeUpdate(std::shared_ptr<ZoneFile> update, bool replace) {
   if (file_id_ != update->GetID())
     return Status::Corruption("ZoneFile update", "ID missmatch");
 
-  Rename(update->GetFilename());
   SetFileSize(update->GetFileSize());
   SetWriteLifeTimeHint(update->GetWriteLifeTimeHint());
   SetFileModificationTime(update->GetFileModificationTime());
@@ -215,11 +221,13 @@ Status ZoneFile::MergeUpdate(std::shared_ptr<ZoneFile> update, bool replace) {
   is_sparse_ = update->IsSparse();
   MetadataSynced();
 
+  linkfiles_.clear();
+  for (const auto& name : update->GetLinkFiles()) linkfiles_.push_back(name);
+
   return Status::OK();
 }
 
-ZoneFile::ZoneFile(ZonedBlockDevice* zbd, std::string filename,
-                   uint64_t file_id)
+ZoneFile::ZoneFile(ZonedBlockDevice* zbd, uint64_t file_id)
     : zbd_(zbd),
       active_zone_(NULL),
       extent_start_(NO_EXTENT),
@@ -227,13 +235,11 @@ ZoneFile::ZoneFile(ZonedBlockDevice* zbd, std::string filename,
       lifetime_(Env::WLTH_NOT_SET),
       io_type_(IOType::kUnknown),
       fileSize(0),
-      filename_(filename),
       file_id_(file_id),
       nr_synced_extents_(0),
       m_time_(0) {}
 
-std::string ZoneFile::GetFilename() { return filename_; }
-void ZoneFile::Rename(std::string name) { filename_ = name; }
+std::string ZoneFile::GetFilename() { return linkfiles_[0]; }
 time_t ZoneFile::GetFileModificationTime() { return m_time_; }
 
 uint64_t ZoneFile::GetFileSize() { return fileSize; }
@@ -703,6 +709,32 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
 
   WriteLock lck(this);
   extents_ = new_list;
+}
+
+void ZoneFile::AddLinkName(const std::string& linkf) {
+  linkfiles_.push_back(linkf);
+}
+
+IOStatus ZoneFile::RenameLink(const std::string& src, const std::string& dest) {
+  auto itr = std::find(linkfiles_.begin(), linkfiles_.end(), src);
+  if (itr != linkfiles_.end()) {
+    linkfiles_.erase(itr);
+    linkfiles_.push_back(dest);
+  } else {
+    return IOStatus::IOError("RenameLink: Failed to find the linked file");
+  }
+  return IOStatus::OK();
+}
+
+IOStatus ZoneFile::RemoveLinkName(const std::string& linkf) {
+  assert(GetNrLinks());
+  auto itr = std::find(linkfiles_.begin(), linkfiles_.end(), linkf);
+  if (itr != linkfiles_.end()) {
+    linkfiles_.erase(itr);
+  } else {
+    return IOStatus::IOError("RemoveLinkInfo: Failed to find the link file");
+  }
+  return IOStatus::OK();
 }
 
 IOStatus ZoneFile::SetWriteLifeTimeHint(Env::WriteLifeTimeHint lifetime) {
--- a/fs/io_zenfs.h
+++ b/fs/io_zenfs.h
@@ -55,6 +55,7 @@ class ZoneFile {
   ZonedBlockDevice* zbd_;
 
   std::vector<ZoneExtent*> extents_;
+  std::vector<std::string> linkfiles_;
 
   Zone* active_zone_;
   uint64_t extent_start_ = NO_EXTENT;
@@ -63,7 +64,6 @@ class ZoneFile {
   Env::WriteLifeTimeHint lifetime_;
   IOType io_type_; /* Only used when writing */
   uint64_t fileSize;
-  std::string filename_;
   uint64_t file_id_;
 
   uint32_t nr_synced_extents_ = 0;
@@ -80,8 +80,7 @@ class ZoneFile {
  public:
   static const int SPARSE_HEADER_SIZE = 8;
 
-  explicit ZoneFile(ZonedBlockDevice* zbd, std::string filename,
-                    uint64_t file_id_);
+  explicit ZoneFile(ZonedBlockDevice* zbd, uint64_t file_id_);
 
   virtual ~ZoneFile();
 
@@ -96,7 +95,6 @@ class ZoneFile {
   IOStatus SetWriteLifeTimeHint(Env::WriteLifeTimeHint lifetime);
   void SetIOType(IOType io_type);
   std::string GetFilename();
-  void Rename(std::string name);
   time_t GetFileModificationTime();
   void SetFileModificationTime(time_t mt);
   uint64_t GetFileSize();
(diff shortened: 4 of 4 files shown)


aravind-wdc:
This also fixes the issue WDCZEN22 reported by Percona, where even when direct read/write was specified, kernel buffer usage was continuously increasing (until all memory was used up, as reported by vmstat) when running in bulk load mode.

In MyRocks, when bulk-load mode is used to create/fill the MySQL database, the DB is ingesting external files into the database, rocksdb does this by creating a link to the external existing sst files. If linking does not work or is not supported by the underlying file system (as in case of ZenFS) , rocksdb falls back on copying the file. However, during this copy it ignores any direct-read/direct-write directives given by the user, and does the copy in the buffered mode, causing continuous increase in the kernel buffer usage.
Also due to the copy, in bulk load mode, to write a DB of size N bytes, we end up doing extra N bytes reads and extra N bytes writes, causing more space amplification.

This patch adds the LinkFile api support, which basically allows rocksdb to create a link instead of doing the copy, so the copy path is avoided in rocksdb. With this patch, kernel buffer usage did not increase at all.

Also since this avoids the copy, in bulk load mode, It reduces the amount of extra reads and writes done due to copy(almost by 2x times).

Will send another patch to RocksDB to ensure the copy is buffered or direct based on user provided instructions.

fs/io_zenfs.cc
```diff
@@ -215,6 +240,13 @@ Status ZoneFile::MergeUpdate(std::shared_ptr<ZoneFile> update, bool replace) {
   is_sparse_ = update->IsSparse();
   MetadataSynced();
 
+  if (nr_links_ < update->GetNrLinks()) {
```
MaisenbacherD, line 243: Is the check correct, that a ZoneFile update can only contain more links? Could it not happen that a ZoneFile update contains less links and thus we need to remove links that do not exist from `this` ZoneFile?

fs/fs_zenfs.cc
```diff
@@ -481,6 +481,27 @@ std::shared_ptr<ZoneFile> ZenFS::GetFile(std::string fname) {
...
+  std::string record;
+
+  files_.erase(fname);
+  zoneFile->RemoveLinkInfo(fname);
+  EncodeFileLinkDeletionTo(zoneFile, &record, fname);
+  s = PersistRecord(record);
+  if (!s.ok()) {
+    /* Failed to sync, return to a consistent state */
+    zoneFile->SetLinkInfo(fname.c_str());
+    zoneFile->SetNrLinks(zoneFile->GetNrLinks() + 1);
```
MaisenbacherD, line 499: Why is it necessary to increment the `nr_links_ one` more time? `SetLinkInfo` should take care of it or not?

fs/fs_zenfs.cc
```diff
@@ -920,6 +1001,36 @@ Status ZenFS::DecodeFileDeletionFrom(Slice* input) {
...
+  if (files_.find(fileName) == files_.end())
+    return Status::Corruption("Zone file link deletion: no such file");
+
+  std::shared_ptr<ZoneFile> zoneFile = files_[fileName];
+  if (zoneFile->GetID() != fileID)
+    return Status::Corruption("Zone file link deletion: file ID missmatch");
+
+  zoneFile->RemoveLinkInfo(fileName);
+  files_.erase(fileName);
+  zoneFile->SetNrLinks(nrLinks);
```
MaisenbacherD, line 1029: I would say `RemoveLinkInfo` should update the `nr_links_` correctly and we should not overwrite that value with `SetNrLinks(nrLinks)`. We could check at this point `if (nrLinks == zoneFile->GetNrLinks())`.

fs/io_zenfs.h
```diff
@@ -140,6 +142,11 @@ class ZoneFile {
   IOStatus Recover();
 
   void ReplaceExtentList(std::vector<ZoneExtent*> new_list);
+  IOStatus SetLinkInfo(std::string linkfile);
+  IOStatus RemoveLinkInfo(std::string linkfile);
+  uint32_t GetNrLinks() { return nr_links_; }
+  void SetNrLinks(uint32_t nrLinks) { nr_links_ = nrLinks; }
```
MaisenbacherD, line 148: I think the `SetNrLinks` method should not exist because `nr_links_` is implicitly set by `SetLinkInfo` and `RemoveLinkInfo`. (Also see comments on `SetNrLinks` references)

MaisenbacherD (commented):
Had some smaller comments. Otherwise it looks good to me :)

fs/fs_zenfs.cc
```diff
@@ -800,6 +827,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+  Debug(logger_, "LinkFile: %s to %s\n", src.c_str(), dst.c_str());
+
+  /* Return if specified link already exists for the src file */
+  dst_file = GetFile(dst);
+  if (dst_file != nullptr)
+    return IOStatus::InvalidArgument("Failed to create link, link exists");
+
+  src_file = GetFile(src);
+  if (src_file != nullptr) {
+    src_file->SetLinkInfo(dst.c_str());
```
MaisenbacherD, line 845: Probalby does not break something:
Can we move `src_file->SetLinkInfo(dst.c_str());` also in the section where `file_mtx_` is locked? `DeleteFile` also has the `file_mtx_` before it modifies links.

fs/fs_zenfs.h
```diff
@@ -170,11 +171,15 @@ class ZenFS : public FileSystemWrapper {
...
                             std::string* output);
+  void EncodeFileLinkDeletionTo(std::shared_ptr<ZoneFile> zoneFile,
+                                std::string* output, std::string linkf);
 
   Status DecodeSnapshotFrom(Slice* input);
   Status DecodeFileUpdateFrom(Slice* slice, bool replace = false);
   Status DecodeFileDeletionFrom(Slice* slice);
+  Status DecodeFileLinkDeletionFrom(Slice* slice);
 
+  IOStatus DeleteFileLink(std::string fname);
```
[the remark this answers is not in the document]
aravind-wdc: Right, removed it.

fs/fs_zenfs.cc
aravind-wdc: Yes, you are right. SetNrLinks()  is not needed here.

fs/fs_zenfs.cc
aravind-wdc: SetLinkInfo() does get the zonefile lock before updating, so should be ok here. The file_mtx_ lock is for the file system, so i did not think it was needed here.  I will move after the lock is held.

fs/io_zenfs.cc
aravind-wdc: Yes it is correct.
When a link is added we encode it as an Update, so when decoding updates, we do a  merging of new updates, so we will be adding a new updated link information to the current snapshot. 

When a link is deleted it is encoded as a FileLinkDeletion operation, it gets decoded at DecodeFileLinkDeletionFrom() , here the file is removed from files_ map and  linkfiles_ vector and links counts are accounted.

fs/fs_zenfs.cc
aravind-wdc: You are right. SetNrLinks() is redundant at this point, will remove it.

fs/io_zenfs.h
aravind-wdc: You are right, SetNrLinks() is redundant at this point, will remove it,

aravind-wdc:
> Had some smaller comments. Otherwise it looks good to me :)

Addressed all your comments, pushed the relevant changes.

fs/io_zenfs.cc
```diff
@@ -705,6 +737,38 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
   extents_ = new_list;
 }
 
+IOStatus ZoneFile::SetLinkInfo(std::string linkf) {
+  WriteLock lck(this);
```
yhr, line 741: We're holding the files_ mutex, why do we need this lock?

yhr (changes requested):
Could you change the status from draft to ready for review? Then CI can do its thing.

Looks alright over all(one comment on the code as it is), but handling renames of linked files seem to be missing. Right now a rename of a linked file would change the name of the original file, right?

aravind-wdc:
> but handling renames of linked files seem to be missing. Right now a rename of a linked file would change the name of the original file, right?

When RenameFile is called, it removes the said filename entry from files_ map and it changes the field filename_ of the object and inserts the new-filename to files_ map, , but the object is still the same (and it is not destroyed) and it still retains the linked files information in the object. So the newfilename->linkedfiles_ would still have the information of the earlier linked files if any. 

For ex: If file "abc.sst" had 2 links, L1-abc.sst, L2-abc.sst, these 2 link entries are stored in the linkfiles_ vector of the object of abc.sst, renaming abc.sst will change filename_ field to say xyz.sst. Now the new file xyz.sst will still be pointing to the old object which has the linkfiles information.

However, in RenameFile(src-file, dest-file, ... ) 
if the dest file is present, zenfs deletes it and then renames the source file as dest file,  in this case if the dest file had any hard links, they will continue to live on and on disk space is not freed, but this is inline with posix behavior. The user needs to delete the links also but user may be doing it intentionally also.

yhr:
> > but handling renames of linked files seem to be missing. Right now a rename of a linked file would change the name of the original file, right?
> 
> When RenameFile is called, it removes the said filename entry from files_ map and it changes the field filename_ of the object and inserts the new-filename to files_ map, , but the object is still the same (and it is not destroyed) and it still retains the linked files information in the object. So the newfilename->linkedfiles_ would still have the information of the earlier linked files if any.
> 
> For ex: If file "abc.sst" had 2 links, L1-abc.sst, L2-abc.sst, these 2 link entries are stored in the linkfiles_ vector of the object of abc.sst, renaming abc.sst will change filename_ field to say xyz.sst. Now the new file xyz.sst will still be pointing to the old object which has the linkfiles information.
> 
> However, in RenameFile(src-file, dest-file, ... ) if the dest file is present, zenfs deletes it and then renames the source file as dest file, in this case if the dest file had any hard links, they will continue to live on and on disk space is not freed, but this is inline with posix behavior. The user needs to delete the links also but user may be doing it intentionally also.

But if you rename a linked filename, lets say "b" . If b is linked to the originally created file a, and you rename b then a is renamed, right?

yhr:
> > > but handling renames of linked files seem to be missing. Right now a rename of a linked file would change the name of the original file, right?
> > 
> > 
> > When RenameFile is called, it removes the said filename entry from files_ map and it changes the field filename_ of the object and inserts the new-filename to files_ map, , but the object is still the same (and it is not destroyed) and it still retains the linked files information in the object. So the newfilename->linkedfiles_ would still have the information of the earlier linked files if any.
> > For ex: If file "abc.sst" had 2 links, L1-abc.sst, L2-abc.sst, these 2 link entries are stored in the linkfiles_ vector of the object of abc.sst, renaming abc.sst will change filename_ field to say xyz.sst. Now the new file xyz.sst will still be pointing to the old object which has the linkfiles information.
> > However, in RenameFile(src-file, dest-file, ... ) if the dest file is present, zenfs deletes it and then renames the source file as dest file, in this case if the dest file had any hard links, they will continue to live on and on disk space is not freed, but this is inline with posix behavior. The user needs to delete the links also but user may be doing it intentionally also.
> 
> But if you rename a linked filename, lets say "b" . If b is linked to the originally created file a, and you rename b then a is renamed, right?

this is what is done in RenameFile:

´´´
    files_mtx_.lock();
    files_.erase(source_path);
    source_file->Rename(dest_path);
    files_.insert(std::make_pair(dest_path, source_file));
    files_mtx_.unlock();
´´´
So in case b is a hard link, the file list will be updated with the new name, but the zone file name(a) will be renamed to the new b name. Right?

aravind-wdc:
> But if you rename a linked filename, lets say "b" . If b is linked to the originally created file a, and you rename b then a is renamed, right?

Ah! I see the case you are pointing out. If we rename a link, on a subsequent mount, the original file will be renamed.
I will update the patch to handle this case.

yhr (changes requested):
I have one open comment - why do we need the write locks inside the new methods (WriteLock ) lck(this);)

fs/io_zenfs.cc
aravind-wdc: Yes, we do not this, holding the files_ mutex should be sufficient. Removed it.

fs/fs_zenfs.cc
```diff
@@ -879,6 +963,11 @@ Status ZenFS::DecodeSnapshotFrom(Slice* input) {
     files_.insert(std::make_pair(zoneFile->GetFilename(), zoneFile));
     if (zoneFile->GetID() >= next_file_id_)
       next_file_id_ = zoneFile->GetID() + 1;
+
+    for (uint32_t i = 0; i < zoneFile->GetNrLinks(); i++) {
+      std::shared_ptr<ZoneFile> z = zoneFile;
+      files_.insert(std::make_pair(z->GetLinkFilename(i), z));
```
metaspace, line 969:
```suggestion
      files_.insert(std::make_pair(z->GetLinkFilename(i), zoneFile));
```
I think there is no reason to make a temporary `shared_ptr` here.

metaspace (changes requested):
I think it would be a better approach to *not* distinguish between regular files and hard links. Instead, create files with just 1 hard link by default, and add more when the link API is invoked. That would reduce code complexity in a number of places, since we do not have to consider whether a file is a regular file or a link.

In general I think we should drop the inline comments when they add no clear value and the code is readable. They can end up doing more bad than god if not properly maintained.

fs/fs_zenfs.cc
aravind-wdc: Yeah, removed the temp variable.

fs/fs_zenfs.cc
```diff
@@ -845,6 +919,11 @@ Status ZenFS::DecodeFileUpdateFrom(Slice* slice, bool replace) {
     if (id == zFile->GetID()) {
       std::string oldName = zFile->GetFilename();
 
+      /* Remove all links from files_, get the updated linkfiles */
+      for (uint32_t i = 0; i < zFile->GetNrLinks(); i++) {
+        if (files_.find(zFile->GetLinkFilename(i)) != files_.end())
```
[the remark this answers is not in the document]
aravind-wdc: Good suggestion, now, returns error if file not found

fs/io_zenfs.h
```diff
@@ -71,6 +72,7 @@ class ZoneFile {
   time_t m_time_;
   bool is_sparse_ = false;
   bool is_deleted_ = false;
+  uint32_t nr_links_ = 0; /* refers to hardlinks */
```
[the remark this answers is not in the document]
aravind-wdc: Removed nr_links_ field

fs/io_zenfs.cc
```diff
@@ -99,6 +101,14 @@ void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
   if (is_sparse_) {
     PutFixed32(output, kIsSparse);
   }
+
+  PutFixed32(output, kNrLinks);
```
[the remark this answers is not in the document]
aravind-wdc: Not needed, removed the nr_links_ field

fs/fs_zenfs.cc
```diff
@@ -895,6 +984,18 @@ void ZenFS::EncodeFileDeletionTo(std::shared_ptr<ZoneFile> zoneFile,
   PutLengthPrefixedSlice(output, Slice(file_string));
 }
 
+void ZenFS::EncodeFileLinkDeletionTo(std::shared_ptr<ZoneFile> zoneFile,
+                                     std::string* output, std::string linkf) {
+  std::string file_string;
+
+  PutFixed64(&file_string, zoneFile->GetID());
+  PutFixed32(&file_string, zoneFile->GetNrLinks());
```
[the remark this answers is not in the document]
aravind-wdc: Not needed, removed the nr_links_ field

fs/io_zenfs.cc
```diff
@@ -705,6 +737,47 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
   extents_ = new_list;
 }
 
+IOStatus ZoneFile::SetLinkInfo(std::string linkf) {
```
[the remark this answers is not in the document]
aravind-wdc: Ok.

fs/fs_zenfs.cc
```diff
@@ -800,6 +842,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+  if (src_file != nullptr) {
+    files_mtx_.lock();
+    src_file->SetLinkInfo(dst.c_str());
+    files_.insert(std::make_pair(dst.c_str(), src_file));
+    files_mtx_.unlock();
+    s = SyncFileMetadata(src_file);
+    if (!s.ok()) {
+      /* Failed to persist the linkedfile, roll back */
+      std::lock_guard<std::mutex> lock(files_mtx_);
+      s = src_file->RemoveLinkInfo(dst.c_str());
```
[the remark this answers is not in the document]
aravind-wdc: Yes it is not needed, I think I saw couple of places where it was passed like this and followed it, removed now.

fs/fs_zenfs.cc
```diff
@@ -800,6 +842,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+
+  /* Return if specified link already exists for the src file */
+  dst_file = GetFile(dst);
+  if (dst_file != nullptr)
+    return IOStatus::InvalidArgument("Failed to create link, link exists");
+
+  src_file = GetFile(src);
+  if (src_file != nullptr) {
+    files_mtx_.lock();
+    src_file->SetLinkInfo(dst.c_str());
```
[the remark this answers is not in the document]
aravind-wdc: Changed the function signature.

fs/fs_zenfs.cc
```diff
@@ -800,6 +842,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+    files_.insert(std::make_pair(dst.c_str(), src_file));
+    files_mtx_.unlock();
+    s = SyncFileMetadata(src_file);
+    if (!s.ok()) {
+      /* Failed to persist the linkedfile, roll back */
```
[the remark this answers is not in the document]
aravind-wdc: Done

fs/io_zenfs.cc
```diff
@@ -215,6 +240,13 @@ Status ZoneFile::MergeUpdate(std::shared_ptr<ZoneFile> update, bool replace) {
   is_sparse_ = update->IsSparse();
   MetadataSynced();
 
+  /* update the linkfiles_ with the updated list */
```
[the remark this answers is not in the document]
aravind-wdc: I think this is a bit helpful in understanding what is happening..

fs/io_zenfs.cc
```diff
@@ -705,6 +737,47 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
   extents_ = new_list;
 }
 
+IOStatus ZoneFile::SetLinkInfo(std::string linkf) {
+  linkfiles_.push_back(linkf.c_str());
+  nr_links_++;
```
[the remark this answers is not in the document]
aravind-wdc: Removed nr_links_, using size()
I was kind of double minded in retaining it, I was thinking nr_links_ could be useful in future, but linkfiles_.size() should serve the purpose. Removed the field entirely.

fs/fs_zenfs.cc
```diff
@@ -800,6 +842,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+IOStatus ZenFS::LinkFile(const std::string& src, const std::string& dst,
+                         const IOOptions& options, IODebugContext* dbg) {
+  std::shared_ptr<ZoneFile> src_file(nullptr);
+  std::shared_ptr<ZoneFile> dst_file(nullptr);
+  IOStatus s;
+
+  Debug(logger_, "LinkFile: %s to %s\n", src.c_str(), dst.c_str());
+
+  /* Return if specified link already exists for the src file */
+  dst_file = GetFile(dst);
```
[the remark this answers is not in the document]
aravind-wdc: Yeah, Fixed this.

fs/fs_zenfs.cc
```diff
@@ -800,6 +842,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+  if (dst_file != nullptr)
+    return IOStatus::InvalidArgument("Failed to create link, link exists");
```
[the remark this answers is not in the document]
aravind-wdc: Ok...

fs/fs_zenfs.cc
```diff
@@ -800,6 +842,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
 
+IOStatus ZenFS::LinkFile(const std::string& src, const std::string& dst,
+                         const IOOptions& options, IODebugContext* dbg) {
+  std::shared_ptr<ZoneFile> src_file(nullptr);
+  std::shared_ptr<ZoneFile> dst_file(nullptr);
+  IOStatus s;
+
+  Debug(logger_, "LinkFile: %s to %s\n", src.c_str(), dst.c_str());
+
+  /* Return if specified link already exists for the src file */
```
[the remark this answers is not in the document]
aravind-wdc: Removed the comment.

fs/fs_zenfs.cc
```diff
@@ -491,6 +511,12 @@ IOStatus ZenFS::DeleteFile(std::string fname) {
     if (zoneFile != nullptr) {
       std::string record;
 
+      if (zoneFile->GetNrLinks()) {
```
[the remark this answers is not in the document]
aravind-wdc: > I think it would be a better approach to _not_ distinguish between regular files and hard links. Instead, create files with just 1 hard link by default, and add more when the link API is invoked. That would reduce code complexity in a number of places, since we do not have to consider whether a file is a regular file or a link.

We are not distinguishing between regular links and hard links, the links also are listed by files_ list as individual files, but the same ZoneFile object backs all the hard linked files. So that any data/metadata update to either the source file or hard links is reflected in all the files.  This is far more convenient than maintaining each hard link as separate file with it's own ZoneFile object , as this would mean updating all linked files whenever the source file data/metadata changes, this also would require to maintain some kind of a 2-way link from source file to linked files to update changes. 


> For readability

Ok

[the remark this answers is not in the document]
metaspace: > We are not distinguishing between regular links and hard links, the links also are listed by files_ list as individual files, but the same ZoneFile object backs all the hard linked files. So that any data/metadata update to either the source file or hard links is reflected in all the files. This is far more convenient than maintaining each hard link as separate file with it's own ZoneFile object , as this would mean updating all linked files whenever the source file data/metadata changes, this also would require to maintain some kind of a 2-way link from source file to linked files to update changes.

I did not mean having a `ZoneFile` object for each link. Just, when we allocate the `ZoneFile`, insert the first path of the file into the `ZoneFile->linkfiles_`, so that `ZoneFile->GetNrLinks() >=1` is invariant. If we do that, we can simplify the code in a few places.

fs/io_zenfs.cc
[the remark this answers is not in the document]
metaspace: In that case, I would suggest factoring out the code that the comment relates to in a function with a descriptive name
```c++
...

  UpdateLinkfilesWithUpdatedList(update); 

  return Status::OK();
}

void ZoneFile::UpdateLinkfilesWithUpdatedList(std::shared_ptr<ZoneFile> &update) {
  linkfiles_.clear();
  for (uint32_t i = 0; i < update->GetNrLinks(); i++) {
    linkfiles_.push_back(update->GetLinkFilename(i));
  }
}
```

fs/fs_zenfs.cc
```diff
@@ -800,6 +841,38 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+  dst_file = GetFileInternal(dst);
+  if (dst_file != nullptr) {
+    files_mtx_.unlock();
+    return IOStatus::InvalidArgument("Failed to create link, target exists");
+  }
+  src_file = GetFileInternal(src);
+  if (src_file != nullptr) {
+    src_file->SetLinkInfo(dst);
+    files_.insert(std::make_pair(dst, src_file));
+    files_mtx_.unlock();
```
metaspace, line 862: We have to hold the lock for the entire duration of the method, or we will race on `src_file`. If the metadata sync fails, the error path may not clean  up properly during a race. @MaisenbacherD is fixing this in #160 - [here](https://github.com/westerndigitalcorporation/zenfs/pull/160/files#diff-b12ba0b7fd2feee53dba42938e4031b32be21c71dc948ba759844c4c49be0a93R440-R467).

I would suggest to wait for that patch set to get merged and then rebase on top of those changes to use the non locking sync method. Then use a `lock_guard` that is dropped before `s = target()->LinkFile(ToAuxPath(src), ToAuxPath(dst), options, dbg);` by returning from the `if` branch.

metaspace (changes requested):

fs/fs_zenfs.cc
aravind-wdc: I can rebase on top of it, but if that is going to take more time, we can merge this and I will submit another patch on top of #160 , fixing the lock issue. 
@yhr thoughts ?

fs/io_zenfs.cc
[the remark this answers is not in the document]
aravind-wdc: Why ? 
If keeping the comment is not good  because it may not be maintained properly in future, same holds true for making the comment into a function name, as the function could be modified in future to do more/less things.

fs/fs_zenfs.cc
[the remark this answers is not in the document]
aravind-wdc: I am not sure if it simplifies more. If it is part of linkfiles_ then we would need to ensure we don't try to re-insert it into 
files_ map, during decoding/encoding the updates.  It would add more corner cases in my opinion.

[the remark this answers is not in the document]
metaspace: This is the decode update now, from [here](https://github.com/westerndigitalcorporation/zenfs/pull/156/commits/7ff22e5200ee5c98158bd2a2e9b816c4b4a8ff76#diff-b12ba0b7fd2feee53dba42938e4031b32be21c71dc948ba759844c4c49be0a93R915-R941):
```c++
/* Remove all links from files_, get the updated linkfiles */
      for (uint32_t i = 0; i < zFile->GetNrLinks(); i++) {
        if (files_.find(zFile->GetLinkFilename(i)) != files_.end())
          files_.erase(zFile->GetLinkFilename(i));
        else
          return Status::Corruption("DecodeFileUpdateFrom: missing link file");
      }
      s = zFile->MergeUpdate(update, replace);
      update.reset();

      if (!s.ok()) return s;

      if (zFile->GetFilename() != oldName) {
        files_.erase(oldName);
        files_.insert(std::make_pair(zFile->GetFilename(), zFile));
      }

      for (uint32_t i = 0; i < zFile->GetNrLinks(); i++) {
        std::shared_ptr<ZoneFile> z = zFile;
        files_.insert(std::make_pair(zFile->GetLinkFilename(i), z));
      }
```
That could be changed to
```c++
/* Remove all links from files_, get the updated linkfiles */
      for (uint32_t i = 0; i < zFile->GetNrLinks(); i++) {
        if (files_.find(zFile->GetLinkFilename(i)) != files_.end())
          files_.erase(zFile->GetLinkFilename(i));
        else
          return Status::Corruption("DecodeFileUpdateFrom: missing link file");
      }
      s = zFile->MergeUpdate(update, replace);
      update.reset();

      if (!s.ok()) return s;

      for (uint32_t i = 0; i < zFile->GetNrLinks(); i++) {
        std::shared_ptr<ZoneFile> z = zFile;
        files_.insert(std::make_pair(zFile->GetLinkFilename(i), z));
      }
```
The first loop removes all links including the first _true_ name because that one is not special. The second loop would insert all links, including the first _true_ name. I don't see how that is more complex.

In the same way, we can completely drop `kFileName` from the on disk format, so encode/decode on disk format would be more simple.

fs/fs_zenfs.cc
```diff
@@ -855,6 +935,11 @@ Status ZenFS::DecodeFileUpdateFrom(Slice* slice, bool replace) {
         files_.insert(std::make_pair(zFile->GetFilename(), zFile));
       }
 
+      for (uint32_t i = 0; i < zFile->GetNrLinks(); i++) {
+        std::shared_ptr<ZoneFile> z = zFile;
+        files_.insert(std::make_pair(zFile->GetLinkFilename(i), z));
```
metaspace, line 940:
```suggestion
        files_.insert(std::make_pair(zFile->GetLinkFilename(i), zFile));
```

fs/fs_zenfs.cc
metaspace: I think it is better to wait than to merge a concurrency bug :shrug:

fs/io_zenfs.cc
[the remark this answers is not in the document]
metaspace: Because, if you move the function invocation around, the message of what it does automatically moves with the invocation and the lines of code that the comment/function name describes. If you have a comment attached to a few lines of code, it can be unclear how much of that code the comment refers to, and moving the code with the comment can be difficult. When refactoring the code at a later stage, the comment may move away from the thing it describes, and then at an even later point in time, someone may forget to update the comment, because it is no longer attached to the code that it actually describes.

It is true that the function name that describe the function content can also rot, but an inline comment that covers some following code rots much faster and is more difficult to maintain.

For deeper explanation check out the comment chapter in [Clean Code by Robert C. Martin](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882) or [this writeup](https://medium.com/@bpnorlander/stop-writing-code-comments-28fef5272752) on medium.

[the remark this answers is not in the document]
aravind-wdc: Trashing any comment as bad coding practice is hardly the right approach. I will remove the comment, but I don't agree that this is a universal rule.

fs/fs_zenfs.cc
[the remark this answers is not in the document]
aravind-wdc: Yeah, this is another way of implementing the api.
Basically we remove the kFileName encoding and decoding and treat filenames as same as links do the filename encoding/decoding as part of linkfiles_ encoding and decoding. 

Pros is that it helps reducing few lines of code. 
Cons are that the filename_ processing also happens a part of linkfiles_ and it will not be straight forward for someone new looking at the code. 
I would still prefer keep them seperate, but we can discuss and take a final call.

fs/fs_zenfs.cc
aravind-wdc: Yeah, I can rebase on Dennis's patch after it is pulled and resubmit.

fs/io_zenfs.cc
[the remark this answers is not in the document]
metaspace: > Trashing any comment as bad coding practice is hardly the right approach.

Let's agree to disagree on that, for in-line comments at least. For API docs it's a different story.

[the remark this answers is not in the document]
metaspace: > I have also observed that people tend to update comments in and around the area that they modify and function headers generally tend to rot as more people add/change functionality of a function.

This is completely true, it happens. But I maintain that inline comments rot way faster than descriptive function names.

> Large scale refactors are frowned upon as the program matures.

We _should_ refactor our code _all the time_ to keep structure clean. If the code is difficult to refactor, we already failed, and we should try to correct that error. Refactoring is not to be avoided, it is to be encouraged. If we choose to break structure because we are afraid that we might break function in the process, then we have a serious problem (maybe we lack proper test coverage?).

> When someone has to fix a priority bug in an area that one is not familiar with, in short time, comments really help. 

If the comments are required to understand the code, we _already failed_ and we should try to fix that rather than make up for it with more comments. **Most of the time there is really no excuse for writing clean code.** If we stumble upon messy code when fixing a bug, we should clean up and leave the code in better shape than when we found it.

>  Infact, within fw engineering, we have tools that measure code/comments ratios.

If the goal of running that tool is to keep the amount of inline comments high, I am out of words. Just because an org chooses to follow such a doctrine does _not_ mean that it a wise choice.

> So IMO, not having comments in the code works only in an perfect world where everyone codes and thinks similarly, else things are bound to slowdown as we try to make sense of the code.

Not having comments work fine if people write their code so that intent is clear and code is easy to read. We don't have to think in a similar way for that to be effective. Most of the time inline comments is an attempt to fix our failure to express ourselves in code. I really like the following quote from Clean Code. It is a bit harsh, but I think it explains very well why we write inline comments:

> The proper use of comments is to compensate for our failure to express ourself in code. Note that I used the word failure. I meant it. Comments are always failures.

I totally respect that people have different opinions on the matter. Also, the viewpoints that I try to convey here are not viewpoints that I invented myself. It is viewpoints that I picked up by studying other peoples opinions on the subject. I find that the "write clean code and drop the comments" camp has better arguments.

For instance take this example from the blog post I posted in an earlier comment: ([source](https://gist.github.com/norMNfan/73e8b31505c0221bfcb70b1481139874#file-wronglevelofabstraction-java))
```java
// This function calculates prices, compares to sales
// promotions, checks if prices are valid, then 
// send an email of promotion to user
public void doSomeThings() {

  // Calculate prices
  ...
    ...
    ...
  
  // Compare calculated prices with sales promotions
  ...
    ...
    ...
  
  // Check if calculated prices are valid
  ...
    ...
    ...
  
  // Send promotions to users
  ...
    ...
    ...
}
```
vs this refactor: ([source](https://gist.github.com/norMNfan/9806c3b382dee9cb9d8cd9601f5e0606#file-wronglevelofabstractionrefactored-java))
```java
public void sendPromotionEmailToUsers() {
  calculatePrices();
  compareCalculatedPricesWithSalesPromotions();
  checkIfCalculatedPricesAreValid();
  sendPromotionEmail();
}
```
I think it is very clear that the refactor is easier to read and understand. 

The same argument holds for the code this comment was originally about. It is easier to read if it is properly decomposed with proper naming. Or do  you think this is not the case?

fs/fs_zenfs.cc
aravind-wdc: This is now rebased on concurrency patches.

aravind-wdc:
Updated to handle filename and linked-files-name in same way.

metaspace:
We need to merge #167 and rebase on the result. CI probably will not work before that.

fs/fs_zenfs.cc
metaspace: I think you need to rebase on master since the concurrency patches are already merged.

aravind-wdc: I will wait for the pkg-config to get merged and then rebase on top of it.

fs/io_zenfs.cc
```diff
@@ -55,22 +55,19 @@ void ZoneExtent::EncodeJson(std::ostream& json_stream) {
...
 enum ZoneFileTag : uint32_t {
   kFileID = 1,
-  kFileName = 2,
-  kFileSize = 3,
-  kWriteLifeTimeHint = 4,
-  kExtent = 5,
-  kModificationTime = 6,
-  kActiveExtentStart = 7,
-  kIsSparse = 8,
+  kFileSize = 2,
```
yhr, line 58: Keep kFilename(just mark is as deprecated) and don't change the other constants, we might want to be able to convert beetween on-disk-formats

fs/fs_zenfs.cc
```diff
@@ -939,7 +1025,11 @@ Status ZenFS::DecodeFileDeletionFrom(Slice* input) {
     return Status::Corruption("Zone file deletion: file ID missmatch");
 
   files_.erase(fileName);
-  zoneFile.reset();
+  s = zoneFile->RemoveLinkName(fileName);
+  if (!s.ok())
+    return Status::Corruption("Zone file deletion: file links missmatch");
+
+  if (zoneFile->GetNrLinks() == 1) zoneFile.reset();
```
yhr, line 1032: Hmm, this should be == 0, right?

fs/io_zenfs.cc
```diff
@@ -182,6 +177,16 @@ Status ZoneFile::DecodeFrom(Slice* input) {
...
         break;
+      case kLinkedFilename:
+        if (!GetLengthPrefixedSlice(input, &slice))
+          return Status::Corruption("ZoneFile", "LinkFilename missing");
+
+        if (slice.ToString().length() == 0)
+          return Status::Corruption("ZoneFile", "Zero length Linkfilename");
+
+        linkfiles_.push_back(slice.ToString());
+        filename_ = linkfiles_[0];
```
yhr, line 188: I think we might as well drop filename_ and have GetFilename return linkfiles_[0]

yhr: Or drop GetFilename..

yhr (changes requested):

fs/fs_zenfs.cc
aravind-wdc: No. Since we add the original filename also to linkfiles_[0], GetNrLInks() returns 1 when there are no more hardlinks to that file and can be reset.

fs/io_zenfs.cc
aravind-wdc: > I think we might as well drop filename_ and have GetFilename return linkfiles_[0]

This is better.

fs/io_zenfs.cc
metaspace: > Keep kFilename(just mark is as deprecated) and don't change the other constants, we might want to be able to convert beetween on-disk-formats

This will be in 2.0 that changes on disk format anyway?

fs/fs_zenfs.cc
```diff
@@ -824,6 +831,79 @@ IOStatus ZenFS::RenameFile(const std::string& source_path,
...
+      if (!s.ok()) {
+        s = src_file->RemoveLinkName(lname);
+        if (!s.ok()) return s;
+        files_.erase(lname);
+      }
+    } else {
+      s = target()->LinkFile(ToAuxPath(fname), ToAuxPath(lname), options, dbg);
+    }
+  }
+  return s;
```
metaspace, line 860: We could let go of the lock before going to aux:
```suggestion
  {
    std::lock_guard<std::mutex> lock(files_mtx_);

    if (GetFileNoLock(lname) != nullptr)
      return IOStatus::InvalidArgument("Failed to create link, target exists");

    src_file = GetFileNoLock(fname);
    if (src_file != nullptr) {
      src_file->SetLinkName(lname);
      files_.insert(std::make_pair(lname, src_file));
      s = SyncFileMetadataNoLock(src_file);
      if (!s.ok()) {
        s = src_file->RemoveLinkName(lname);
        if (!s.ok()) return s;
        files_.erase(lname);
      }
      return s;
    } 
  }
  s = target()->LinkFile(ToAuxPath(fname), ToAuxPath(lname), options, dbg);
```

fs/fs_zenfs.cc
```diff
@@ -867,17 +947,20 @@ Status ZenFS::DecodeFileUpdateFrom(Slice* slice, bool replace) {
...
       update.reset();
 
       if (!s.ok()) return s;
 
-      if (zFile->GetFilename() != oldName) {
-        files_.erase(oldName);
-        files_.insert(std::make_pair(zFile->GetFilename(), zFile));
-      }
+      for (uint32_t i = 0; i < zFile->GetNrLinks(); i++)
+        files_.insert(std::make_pair(zFile->GetLinkFilename(i), zFile));
```
metaspace, line 963: Just a thought, what happens if we insert the same name twice? Could happen with corrupt data or error when encoding.

fs/io_zenfs.cc
```diff
@@ -215,25 +220,28 @@ Status ZoneFile::MergeUpdate(std::shared_ptr<ZoneFile> update, bool replace) {
   is_sparse_ = update->IsSparse();
   MetadataSynced();
 
+  linkfiles_.clear();
+  for (uint32_t i = 0; i < update->GetNrLinks(); i++) {
+    linkfiles_.push_back(update->GetLinkFilename(i));
+  }
```
metaspace, line 226: After this change, [this](https://github.com/westerndigitalcorporation/zenfs/pull/156/files#diff-620cbbece5179b69096a6f453042d59dffedcb5672fb0f4b7f343d68e664314aR203) line has no effect and should be removed.

metaspace (changes requested):
I would suggest to remove `ZoneFile::Rename()` as part of this PR.

fs/io_zenfs.cc
```diff
@@ -705,6 +713,36 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
   extents_ = new_list;
 }
 
+void ZoneFile::SetLinkName(const std::string& linkf) {
+  linkfiles_.push_back(linkf);
+}
+
+IOStatus ZoneFile::RenameLink(std::string src, std::string dest) {
+  if (GetFilename() == src) {
```
yhr, line 721: It does not make sense to me to keep rename() treating the first name in the list special. Removing rename(), we can simplify this function..

fs/io_zenfs.cc
```diff
@@ -705,6 +713,36 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
   extents_ = new_list;
 }
 
+void ZoneFile::SetLinkName(const std::string& linkf) {
```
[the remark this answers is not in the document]
yhr: AddLinkName sounds good to me

fs/io_zenfs.cc
```diff
@@ -99,12 +97,17 @@ void ZoneFile::EncodeTo(std::string* output, uint32_t extent_start) {
...
+    PutFixed32(output, kLinkedFilename);
+    PutLengthPrefixedSlice(output, Slice(linkfiles_[i]));
+  }
 }
 
 void ZoneFile::EncodeJson(std::ostream& json_stream) {
   json_stream << "{";
   json_stream << "\"id\":" << file_id_ << ",";
-  json_stream << "\"filename\":\"" << filename_ << "\",";
+  json_stream << "\"filename\":\"" << linkfiles_[0] << "\",";
```
yhr, line 110: Maybe list all the filenames?

fs/fs_zenfs.cc
yhr: We should add a check for that

fs/io_zenfs.cc
aravind-wdc: I did not get what you are referring to, the hyperlink points to a diff rather than a line.

fs/fs_zenfs.cc
aravind-wdc: insert would not overwrite or insert a new one, if the same key already exists in the map.

fs/fs_zenfs.cc
aravind-wdc: Ok we can do that, but  this is still the case with other api's and also it ensures that we do not get into weird race conditions. Do you think holding it affects the performance ?

fs/io_zenfs.cc
```diff
@@ -55,22 +55,20 @@ void ZoneExtent::EncodeJson(std::ostream& json_stream) {
 
 enum ZoneFileTag : uint32_t {
   kFileID = 1,
-  kFileName = 2,
+  kFileName = 2,  // Deprecated
```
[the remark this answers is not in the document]
aravind-wdc: @yhr Are you ok with this ?

fs/io_zenfs.h
```diff
@@ -140,6 +139,11 @@ class ZoneFile {
   IOStatus Recover();
 
   void ReplaceExtentList(std::vector<ZoneExtent*> new_list);
+  void SetLinkName(const std::string& linkfile);
+  IOStatus RemoveLinkName(std::string linkfile);
+  IOStatus RenameLink(std::string src, std::string dest);
+  uint32_t GetNrLinks() { return linkfiles_.size(); }
+  std::string GetLinkFilename(uint32_t idx) { return linkfiles_[idx]; }
```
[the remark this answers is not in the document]
aravind-wdc: GetLinkFilename() can also be used to get the name based on the index.
Do you think the current implementation has any negative performance implications ? If not, I think we can go with it.

fs/io_zenfs.cc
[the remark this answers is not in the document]
aravind-wdc: > Can we rename this one to AddLinkName?
Ok, AddLinkName better encapsulates what is happening..
 
> I would suggest to remove ZoneFile::Rename() as part of this PR.
We can remove it, but I think it is a good way to indicate what is happening instead of just doing 
`linkfiles_[0] = newname`. So it does not hurt to keep it. Lets keep it.

fs/io_zenfs.h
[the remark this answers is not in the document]
yhr: I think this makes a lot of sense

fs/fs_zenfs.cc
aravind-wdc: The existing logic before this patch was to do a zoneFile.reset() during file deletion (I think we have debated on necessity of that, but that debate is not the part of this PR). 
Now, since we have hardlinks, on file deletion we cannot just do a zoneFile.reset() , we need to do it when we reach the last  hardlink (count is one).

yhr (changes requested):

fs/io_zenfs.cc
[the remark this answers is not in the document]
yhr: yep!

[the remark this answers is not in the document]
yhr: Might as well add a comment that it is deperecated in the the new format version, and that it is replaced by the links

fs/io_zenfs.h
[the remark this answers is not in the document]
metaspace: > Do you think the current implementation has any negative performance implications ?

I think performance is going to be the same, but code reads more clearly with the iterator API.

fs/fs_zenfs.cc
metaspace: For performance, I don't think it will matter much in the general case. We are not calling this function often. But if something happened to the `target()` that would cause high latency, we would prevent others form accessing `files_` in the time we are waiting for the target filesystem.

I don't think it will be a race. As far as I can see, the only thing that could happen that we check for `fname` in `files_` and it is not there, so we drop the lock. Next thing someone could `OpenWritableFile()` with the same name, and after that we may proceed to call `target()->LinkFile()` even though the source now exist in ZenFS. But that would happen anyway if `OpenWritableFile()` was interleaved after `target()->LinkFile()`.

fs/fs_zenfs.cc
metaspace: You are right. I was certain that `files_` was a vector of pairs. Never mind then.

fs/io_zenfs.cc
metaspace: Sorry, I thought it retained the line info. I was trying to refer to `Rename()` earlier in the method.

fs/io_zenfs.cc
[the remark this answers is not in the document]
metaspace: `Rename()` is only called in two locations, both of which are redundant. It can just be removed.

fs/io_zenfs.cc
metaspace: Agreed, we can just remove the `if` part of this branch and only keep the `else` part.

fs/fs_zenfs.cc
metaspace: So, before we had a line of code with absolutely no effect. Now we have a more complex line of code with _no effect_. It is completely redundant. It makes the code difficult to read, because nobody can figure out _why_ it is there in the first place. 

What does it do? Nothing. So remove it. It does _not_ help understand anything and it has no functional effect.

If you can tell me why we need to perform a redundant operation on a stack local smart pointer *only in the case that we have exactly on one link to a file after decoding a deletion*, then keep it. Otherwise I think it is a really bad idea.

aravind-wdc: Well, I am aware of the things you have said, as far as the functionality goes.
I don't agree with  `nobody can figure out... ` kind of statements, because it is obvious what is happening.

If a file has say, 3 hard links, and one of them was deleted, then in the decode path, then the corresponding object's shared pointer was getting reset. Reset() in decode path was ok when each file  was an individual entity, but now there could be situations where the deleted file was a hard link(or a original file) and there are still references to the object and the decode was resetting it. So to retain earlier code as is, I added a check do the reset only if it was a last hard link.
 
I was under the impression that you wanted to remove the if condition only. But you want to remove the reset() call also, right ?  That should be ok.

aravind-wdc:
Addressed all latest comments.

yhr (approved):
Looks very neat now.

fs/io_zenfs.cc
```diff
@@ -705,6 +711,32 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
   extents_ = new_list;
 }
 
+void ZoneFile::AddLinkName(const std::string& linkf) {
+  linkfiles_.push_back(linkf);
+}
+
+IOStatus ZoneFile::RenameLink(std::string src, std::string dest) {
```
metaspace, line 718: We should pass arguments by const reference here

fs/io_zenfs.cc
```diff
@@ -705,6 +711,32 @@ void ZoneFile::ReplaceExtentList(std::vector<ZoneExtent*> new_list) {
...
+  if (itr != linkfiles_.end()) {
+    linkfiles_.erase(itr);
+    linkfiles_.push_back(dest);
+  } else {
+    return IOStatus::IOError("RenameLink: Failed to find the linked file");
+  }
+  return IOStatus::OK();
+}
+
+IOStatus ZoneFile::RemoveLinkName(std::string linkf) {
```
metaspace, line 729: Same here, const ref argument.

metaspace (changes requested):
Looks good! Just have one comment about string arguments.

aravind-wdc:
Addressed latest review comments.

PR #160, opened 2022-02-23
Support RocksDB checkpointing

MaisenbacherD:
These commits enable ZenFS to ZenFS checkpointing in RocksDB.
Renaming was fixed such that all ZenFS file names are renamed according to a directory rename.
The `force-rmdir` util command was introduced so that checkpoints can be deleted (which is also needed by the rocksdb.checkpoint MTR test case). 
Addresses jira issue WDCZEN-19.

Reported-by: Yura Sorokin <yura.sorokin@percona.com>

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -692,6 +692,56 @@ IOStatus ZenFS::GetChildren(const std::string& dir, const IOOptions& options,
   return GetChildrenNoLock(dir, options, result, dbg);
 }
 
+/* Must hold files_mtx_ */
+IOStatus ZenFS::DeleteDirRecursiveNoLock(const std::string& d,
+                                         const IOOptions& options,
+                                         IODebugContext* dbg) {
+  std::vector<std::string> children;
+  IOStatus s;
+
+  Debug(logger_, "DeleteDirRecursiveNoLock: %s aux: %s\n", d.c_str(),
+        ToAuxPath(d).c_str());
+
+  s = GetChildrenNoLock(d, options, &children, dbg);
+  if (!s.ok()) {
+    return s;
+  }
+
+  for (const auto& child : children) {
+    std::string file_to_delete =
+        (std::filesystem::path(d) / std::filesystem::path(child)).string();
+    bool is_dir;
+
+    s = IsDirectoryNoLock(file_to_delete, options, &is_dir, dbg);
+    if (!s.ok()) {
+      return s;
+    }
+
+    if (is_dir) {
+      s = DeleteDirRecursiveNoLock(file_to_delete, options, dbg);
+    } else {
+      s = DeleteFileNoLock(file_to_delete, options, dbg);
+    }
+    if (!s.ok()) {
+      return s;
+    }
+  }
+
+  return target()->DeleteDir(ToAuxPath(d), options, dbg);
+}
+
+IOStatus ZenFS::DeleteDirRecursive(const std::string& d,
+                                   const IOOptions& options,
+                                   IODebugContext* dbg) {
+  IOStatus s;
+  {
+    std::lock_guard<std::mutex> lock(files_mtx_);
+    s = DeleteDirRecursiveNoLock(d, options, dbg);
+  }
+  if (s.ok()) s = zbd_->ResetUnusedIOZones();
+  return s;
+}
+
 IOStatus ZenFS::OpenWritableFile(const std::string& fname,
                                  const FileOptions& file_opts,
                                  std::unique_ptr<FSWritableFile>* result,
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -795,6 +795,80 @@ IOStatus ZenFS::GetFileSize(const std::string& f, const IOOptions& options,
 }
 
 /* Must hold files_mtx_ */
+IOStatus ZenFS::RenameChildNoLock(std::string const& source_dir,
+                                  std::string const& dest_dir,
+                                  std::string const& child,
+                                  const IOOptions& options,
+                                  IODebugContext* dbg) {
+  std::string source_child =
+      (std::filesystem::path(source_dir) / std::filesystem::path(child))
+          .string();
+  std::string dest_child =
+      (std::filesystem::path(dest_dir) / std::filesystem::path(child)).string();
+  return RenameFileNoLock(source_child, dest_child, options, dbg);
+}
+
+/* Must hold files_mtx_ */
+IOStatus ZenFS::RollbackAuxDirRenameNoLock(
+    const std::string& source_path, const std::string& dest_path,
+    const std::vector<std::string>& renamed_children, const IOOptions& options,
+    IODebugContext* dbg) {
+  IOStatus s;
+
+  for (const auto& rollback_child : renamed_children) {
+    s = RenameChildNoLock(dest_path, source_path, rollback_child, options, dbg);
+    if (!s.ok()) {
+      return IOStatus::Corruption(
+          "RollbackAuxDirRenameNoLock: Failed to roll back directory rename");
+    }
+  }
+
+  s = target()->RenameFile(ToAuxPath(dest_path), ToAuxPath(source_path),
+                           options, dbg);
+  if (!s.ok()) {
+    return IOStatus::Corruption(
+        "RollbackAuxDirRenameNoLock: Failed to roll back auxiliary path "
+        "renaming");
+  }
+
+  return s;
+}
+
+/* Must hold files_mtx_ */
+IOStatus ZenFS::RenameAuxPathNoLock(const std::string& source_path,
+                                    const std::string& dest_path,
+                                    const IOOptions& options,
+                                    IODebugContext* dbg) {
+  IOStatus s;
+  std::vector<std::string> children;
+  std::vector<std::string> renamed_children;
+
+  s = target()->RenameFile(ToAuxPath(source_path), ToAuxPath(dest_path),
+                           options, dbg);
+  if (!s.ok()) {
+    return s;
+  }
+
+  GetZenFSChildrenNoLock(source_path, true, &children);
+
+  for (const auto& child : children) {
+    s = RenameChildNoLock(source_path, dest_path, child, options, dbg);
+    if (!s.ok()) {
+      IOStatus failed_rename = s;
+      s = RollbackAuxDirRenameNoLock(source_path, dest_path, renamed_children,
+                                     options, dbg);
+      if (!s.ok()) {
+        return s;
+      }
+      return failed_rename;
+    }
+    renamed_children.push_back(child);
+  }
+
+  return s;
+}
+
+/* Must hold files_mtx_ */
 IOStatus ZenFS::RenameFileNoLock(const std::string& source_path,
                                  const std::string& dest_path,
                                  const IOOptions& options,
@@ -831,8 +905,7 @@ IOStatus ZenFS::RenameFileNoLock(const std::string& source_path,
       files_.insert(std::make_pair(source_path, source_file));
     }
   } else {
-    s = target()->RenameFile(ToAuxPath(source_path), ToAuxPath(dest_path),
-                             options, dbg);
+    s = RenameAuxPathNoLock(source_path, dest_path, options, dbg);
   }
 
   return s;
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -15,6 +15,7 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <filesystem>
 #include <sstream>
 #include <utility>
 #include <vector>
@@ -621,11 +622,44 @@ IOStatus ZenFS::ReopenWritableFile(const std::string& fname,
 }
 
 /* Must hold files_mtx_ */
+void ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
+                                   bool include_grandchildren,
+                                   std::vector<std::string>* result) {
+  auto path_as_string_with_separator_at_end =
+      [](std::filesystem::path const& path) {
+        auto with_sep = path / "";
+        return with_sep.lexically_normal().string();
+      };
+
+  auto string_starts_with = [](std::string const& string,
+                               std::string const& needle) {
+    return string.rfind(needle, 0) == 0;
+  };
+
+  std::string dir_with_terminating_seperator =
+      path_as_string_with_separator_at_end(std::filesystem::path(dir));
+
+  for (auto const& it : files_) {
+    std::filesystem::path file_path(it.first);
+    assert(file_path.has_filename());
+
+    std::string file_dir =
+        path_as_string_with_separator_at_end(file_path.parent_path());
+
+    if (string_starts_with(file_dir, dir_with_terminating_seperator)) {
+      if (include_grandchildren ||
+          file_dir.length() == dir_with_terminating_seperator.length()) {
+        result->push_back(file_path.filename().string());
+      }
+    }
+  }
+}
+
+/* Must hold files_mtx_ */
 IOStatus ZenFS::GetChildrenNoLock(const std::string& dir,
                                   const IOOptions& options,
                                   std::vector<std::string>* result,
                                   IODebugContext* dbg) {
-  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   std::vector<std::string> auxfiles;
   IOStatus s;
 
@@ -646,22 +680,7 @@ IOStatus ZenFS::GetChildrenNoLock(const std::string& dir,
     if (f != "." && f != "..") result->push_back(f);
   }
 
-  for (it = files_.begin(); it != files_.end(); it++) {
-    std::string fname = it->first;
-    if (fname.rfind(dir, 0) == 0) {
-      if (dir.back() == '/') {
-        fname.erase(0, dir.length());
-      } else {
-        if (dir.size() > 0) {
-          fname.erase(0, dir.length() + 1);
-        }
-      }
-      // Don't report grandchildren
-      if (fname.find("/") == std::string::npos) {
-        result->push_back(fname);
-      }
-    }
-  }
+  GetZenFSChildrenNoLock(dir, false, result);
 
   return s;
 }
--- a/fs/fs_zenfs.h
+++ b/fs/fs_zenfs.h
@@ -197,6 +197,10 @@ class ZenFS : public FileSystemWrapper {
   /* Must hold files_mtx_ */
   std::shared_ptr<ZoneFile> GetFileNoLock(std::string fname);
   /* Must hold files_mtx_ */
+  void GetZenFSChildrenNoLock(const std::string& dir,
+                              bool include_grandchildren,
+                              std::vector<std::string>* result);
+  /* Must hold files_mtx_ */
   IOStatus GetChildrenNoLock(const std::string& dir, const IOOptions& options,
                              std::vector<std::string>* result,
                              IODebugContext* dbg);
--- a/fs/fs_zenfs.h
+++ b/fs/fs_zenfs.h
@@ -204,6 +204,24 @@ class ZenFS : public FileSystemWrapper {
   IOStatus GetChildrenNoLock(const std::string& dir, const IOOptions& options,
                              std::vector<std::string>* result,
                              IODebugContext* dbg);
+
+  /* Must hold files_mtx_ */
+  IOStatus RenameChildNoLock(std::string const& source_dir,
+                             std::string const& dest_dir,
+                             std::string const& child, const IOOptions& options,
+                             IODebugContext* dbg);
+
+  /* Must hold files_mtx_ */
+  IOStatus RollbackAuxDirRenameNoLock(
+      const std::string& source_path, const std::string& dest_path,
+      const std::vector<std::string>& renamed_children,
+      const IOOptions& options, IODebugContext* dbg);
+
+  /* Must hold files_mtx_ */
+  IOStatus RenameAuxPathNoLock(const std::string& source_path,
+                               const std::string& dest_path,
+                               const IOOptions& options, IODebugContext* dbg);
+
   /* Must hold files_mtx_ */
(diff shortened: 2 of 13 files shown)


aravind-wdc (commented):
I would suggest splitting the changes in this patch to 2 commits.
One to handle the rename issue.
Second to handle the improvements to RenameFile.
It would be easier to understand the context of the changes.

dennis:
> @dennis Could you also please explain on why this is needed, like the problem we are trying to solve ?

I'd rather not. But fyi, I get notified when you mention my username like that :)

aravind-wdc:
> > @dennis Could you also please explain on why this is needed, like the problem we are trying to solve ?
> 
> I'd rather not. But fyi, I get notified when you mention my username like that :)

Woah!!! 
Sorry, did not know it tag someone not in the PR.
I meant our Dennis :)

MaisenbacherD:
> Could you also please explain on why this is needed, like the problem we are trying to solve ?

Sure :)
We want to be able to create RocksDB checkpoints from within ZenFS, which should be faster than making a backup (given that linking is implemented). After a successful checkpoint was created, a backup can be created while RocksDB is able to continue utilising ZenFS. The checkpoint can be deleted after the backup finished.

We can see in the RocksDB checkpointing implementation (https://github.com/facebook/rocksdb/blob/v6.25.3/utilities/checkpoint/checkpoint_impl.cc#L156) that a temporary directory is first created where all the files that need to be checkpointed are linked/copied into. At the end the temporary directory is renamed to the given checkpoint path. 
ZenFS does implement the concept of directories and rather creates files which have the absolute path (relative from the auxiliary directory) as a filename. Thus we need to rename all ZenFS files that relate to an auxiliary directory that is being renamed.

metaspace (changes requested):

fs/fs_zenfs.cc
```diff
@@ -820,8 +820,39 @@ IOStatus ZenFS::RenameFileInternal(const std::string& source_path,
...
+    std::string dest_seperator = (dest_path.back() != '/') ? "/" : "";
+    std::vector<std::string> children;
+    s = GetChildrenInternal(source_path, options, &children, dbg);
+    if (!s.ok()) {
+      return s;
+    }
+
+    for (const auto& child : children) {
+      s = RenameFileInternal(source_path + source_seperator + child,
+                             dest_path + dest_seperator + child, options, dbg);
```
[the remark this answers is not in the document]
MaisenbacherD: ZenFS does not implement the concept of directories. If you use the `NewDirectory` or `CreateDir` functions of ZenFS you will create a directory in AUX.

fs/fs_zenfs.cc
```diff
@@ -744,13 +789,19 @@ IOStatus ZenFS::DeleteFile(const std::string& fname, const IOOptions& options,
...
 
   return s;
 }
 
+IOStatus ZenFS::DeleteFile(const std::string& fname, const IOOptions& options,
+                           IODebugContext* dbg) {
+  std::lock_guard<std::mutex> lock(files_mtx_);
+  return DeleteFileInternal(fname, options, dbg);
+}
+
```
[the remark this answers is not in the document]
MaisenbacherD: Yes makes sense.

fs/fs_zenfs.cc
```diff
@@ -481,32 +486,47 @@ std::shared_ptr<ZoneFile> ZenFS::GetFile(std::string fname) {
...
+/* Must hold files_mtx_ */
+IOStatus ZenFS::DeleteFileInternal(std::string fname) {
+  IOStatus s = DeleteZoneFileInternal(fname);
+  if (s.ok()) s = zbd_->ResetUnusedIOZones();
+  return s;
+}
+
+IOStatus ZenFS::DeleteFile(std::string fname) {
```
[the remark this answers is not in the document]
MaisenbacherD: Agree, I will refactor the *Internal methods :)

fs/fs_zenfs.cc
```diff
@@ -827,8 +827,39 @@ IOStatus ZenFS::RenameFileNoLock(const std::string& source_path,
...
+      return s;
+    }
+
+    for (const auto& child : children) {
+      s = RenameFileNoLock(source_path + source_seperator + child,
+                           dest_path + dest_seperator + child, options, dbg);
+      if (!s.ok()) {
+        /* roll back renameing */
+        for (const auto& rollback_child : children) {
+          RenameFileNoLock(dest_path + dest_seperator + rollback_child,
```
metaspace, line 888: We are discarding the return status of this call. What would happen if we had an error during rollback? Should we ignore it or should we stop and bail?

fs/fs_zenfs.cc
```diff
@@ -827,8 +827,39 @@ IOStatus ZenFS::RenameFileNoLock(const std::string& source_path,
...
+      }
+    }
+
     s = target()->RenameFile(ToAuxPath(source_path), ToAuxPath(dest_path),
                              options, dbg);
+
+    if (!s.ok()) {
+      /* roll back renameing */
+      for (const auto& rollback_child : children) {
+        RenameFileNoLock(dest_path + dest_seperator + rollback_child,
```
metaspace, line 902: Same here about discarding the status.

metaspace (changes requested):

fs/fs_zenfs.cc
MaisenbacherD: Maybe it makes sense to return a `IOStatus::Corruption` at that point?

metaspace: I think that would make sense. Nothing can be trusted at this point, so better to give up and shut down.

yhr: IOStatus::Corruption would be fine

util/zenfs.cc
```diff
@@ -604,6 +637,8 @@ int main(int argc, char **argv) {
     return ROCKSDB_NAMESPACE::zenfs_tool_dump();
   } else if (subcmd == "fs-info") {
     return ROCKSDB_NAMESPACE::zenfs_tool_fsinfo();
+  } else if (subcmd == "force-rmdir") {
```
yhr, line 640: Lets introduce a normal rmdir that would remove a directory if empty, and if it is not empty require the --force flag

yhr (changes requested):

fs/fs_zenfs.cc
MaisenbacherD: IOStatus::Corruption is now being returned in that case.

fs/fs_zenfs.cc
MaisenbacherD: IOStatus::Corruption is now being returned in that case.

util/zenfs.cc
MaisenbacherD: Good point! This feature is now implemented :)

MaisenbacherD:
Rebased the changes onto master, since https://github.com/westerndigitalcorporation/zenfs/pull/163 introduced commits that this PR also contained.

fs/fs_zenfs.cc
```diff
@@ -805,8 +805,49 @@ IOStatus ZenFS::RenameFileNoLock(const std::string& source_path,
...
+          if (!s.ok()) {
+            return IOStatus::Corruption(
+                "ZenFS RenameFileNoLock: Unable to roll back unsuccessful "
+                "renameing.");
+          }
+        }
+        return s;
+      }
+    }
+
```
metaspace, line 883: Since RocksDB moved to c++17 I think it makes sense to use `std::filesystem::path::append` to concatenate components of the paths.

metaspace (changes requested):

fs/fs_zenfs.cc
MaisenbacherD: @percona-ysorokin clarified that percona-server release-8.0.28-19 enables c++17 without the need to switch to RocksDB 7.0.x.

MaisenbacherD:
Now using `std::filesystem::path::append` instead of manual path parsing.

yhr:
As this depends on the new features introduced, I'm checking with bytedance if they would be ok to move to c++17. They are using boost though, and that also has filesystem::path - so we could use that worst case I believe.

MaisenbacherD:
Fixed a bug in `ZenFS::RenameFileNoLock` that @metaspace discovered:
The aux directory needs to be renamed prior to its children. 

Furthermore only the children that are ZenFS files need adjustment because any other children in the renamed aux directory were covered by the posix rename of this directory.

MaisenbacherD:
Rebased onto master

fs/fs_zenfs.cc
```diff
@@ -620,12 +620,38 @@ IOStatus ZenFS::ReopenWritableFile(const std::string& fname,
...
+                                       bool include_grandchildren,
+                                       std::vector<std::string>* result,
+                                       const IOOptions& options,
+                                       IODebugContext* dbg) {
+  IOStatus s;
+  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
+
+  for (it = files_.begin(); it != files_.end(); it++) {
+    std::string fname = it->first;
+    if (fname.rfind(dir, 0) == 0) {
```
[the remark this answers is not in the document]
MaisenbacherD: Maybe we can add something like this:
```suggestion
    if (fname.rfind(std::filesystem::path(dir + "/").lexically_normal().string(), 0) == 0) {
```

fs/fs_zenfs.h
```diff
@@ -359,6 +363,9 @@ class ZenFS : public FileSystemWrapper {
     return target()->DeleteDir(ToAuxPath(d), options, dbg);
   }
 
+  IOStatus ForceDeleteDir(const std::string& d, const IOOptions& options,
+                          IODebugContext* dbg);
```
aravind-wdc, line 367: I think `IODebugContext* dbg` is not needed as it is not a Rocksdb defined API and also we are not using this anywhere to pass back any information.

util/zenfs.cc
```diff
@@ -656,6 +656,49 @@ int zenfs_tool_rename_file() {
...
+
+  std::unique_ptr<ZenFS> zenFS;
+  s = zenfs_mount(zbd, &zenFS, false);
+  if (!s.ok()) {
+    fprintf(stderr, "Failed to mount filesystem, error: %s\n",
+            s.ToString().c_str());
+    return 1;
+  }
+
+  if (FLAGS_force) {
```
aravind-wdc, line 680: At line 38, Should we add "If used with  mkfs or rmdir commands"

fs/fs_zenfs.cc
```diff
@@ -786,6 +787,41 @@ IOStatus ZenFS::GetFileSize(const std::string& f, const IOOptions& options,
...
+    std::string dest_rollback_child_path =
+        (std::filesystem::path(source_path) /
+         std::filesystem::path(rollback_child))
+            .string();
+    s = RenameFileNoLock(source_rollback_child_path, dest_rollback_child_path,
+                         options, dbg);
+    if (!s.ok()) {
+      return IOStatus::Corruption(
+          "ZenFS RenameFileNoLock: Unable to roll back unsuccessful "
+          "ZenFS file renameing.");
```
aravind-wdc, line 859: Typo: renameing -> renaming

fs/fs_zenfs.cc
```diff
@@ -786,6 +787,41 @@ IOStatus ZenFS::GetFileSize(const std::string& f, const IOOptions& options,
...
+    }
+  }
+
+  s = target()->RenameFile(ToAuxPath(dest_path), ToAuxPath(source_path),
+                           options, dbg);
+  if (!s.ok()) {
+    return IOStatus::Corruption(
+        "ZenFS RenameFileNoLock: Unable to roll back auxiliary path "
+        "renameing after unsuccessful ZenFS file renameing.");
```
aravind-wdc, line 868: Error msg feels a bit log, a simplified version will be helpful, I guess ? may be
`RenameFileNoLock: Failed to roll back auxiliary path renaming`

fs/fs_zenfs.cc
```diff
@@ -823,8 +859,40 @@ IOStatus ZenFS::RenameFileNoLock(const std::string& source_path,
       files_.insert(std::make_pair(source_path, source_file));
     }
   } else {
+    std::vector<std::string> children;
```
aravind-wdc, line 911: This function is getting very long, can we move the new logic into a separate function  say, RenameDir() or something more apt, and call that function here.

fs/fs_zenfs.cc
```diff
@@ -620,12 +620,38 @@ IOStatus ZenFS::ReopenWritableFile(const std::string& fname,
...
 }
 
+/* Must hold files_mtx_ */
+IOStatus ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
+                                       bool include_grandchildren,
+                                       std::vector<std::string>* result,
+                                       const IOOptions& options,
+                                       IODebugContext* dbg) {
+  IOStatus s;
+  std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
```
aravind-wdc, line 631: We can remove this line(L630) and directly use:
`for (auto it = files_.begin(); it != files_.end(); it++) {`

aravind-wdc (changes requested):

MaisenbacherD:
Adressed the comments @aravind-wdc had and bumped the RocksDB version for CI so that c++17 is used.

Also refactored `GetZenFSChildrenNoLock` so that we always check if the file name starts with dir plus a file separator.
Otherwise we would find false children if the dir string is a filename (not a directory) that is a substring of an existing directory.

yhr:
Looks allright to me, awaiting aravinds/dennis ok as well.

fs/fs_zenfs.cc
```diff
@@ -620,12 +621,33 @@ IOStatus ZenFS::ReopenWritableFile(const std::string& fname,
...
+          (file_path.parent_path() / "").lexically_normal().string();
+      if (file_dir.rfind(dir_with_terminating_seperator, 0) == 0) {
+        if (include_grandchildren ||
+            file_dir.length() == dir_with_terminating_seperator.length()) {
+          result->push_back(file_path.filename().string());
+        }
+      }
+    }
+  }
+}
```
metaspace, line 644:
```suggestion
void ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
                                   bool include_grandchildren,
                                   std::vector<std::string> &result) {
  auto path_as_string_with_separator_at_end =
      [](std::filesystem::path const& path) {
        auto with_sep = path / "";
        return with_sep.lexically_normal().string();
      };

  auto string_starts_with = [](std::string const& string,
                               std::string const& needle) {
    return string.rfind(needle, 0) == 0;
  };

  std::string dir_with_terminating_seperator =
      path_as_string_with_separator_at_end(std::filesystem::path(dir));

  for (auto const& it : files_) {
    std::filesystem::path file_path(it.first);
    assert(file_path.has_filename());
    // if (!file_path.has_filename()) continue; // <- This should be impossible?

    std::string file_dir =
        path_as_string_with_separator_at_end(file_path.parent_path());

    if (string_starts_with(file_dir, dir_with_terminating_seperator)) {
      if (include_grandchildren ||
          file_dir.length() == dir_with_terminating_seperator.length()) {
        result.push_back(file_path.filename().string());
      }
    }
  }
}
```

Do we need to add the separator at end before comparison? Why can we not compare normalized paths without the terminating separator?

metaspace (changes requested):
Can we change the name of  `ForceDeleteDir` to `DeleteDirRecursive`?

fs/fs_zenfs.cc
MaisenbacherD: I like the refactoring suggestions!

We need the separator at the end before comparison, otherwise we would find false children if the dir string is a filename (not a directory) that is a substring of an existing directory.
Normalising a path without an explicit separator at the end of the path does not add the separator at the end of a directory https://en.cppreference.com/w/cpp/filesystem/path/lexically_normal.

metaspace: We could check that `dir` is a directory in `target()` first, then drop the separator append?

metaspace:
```c++
void ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
                                   bool include_grandchildren,
                                   std::vector<std::string>& result) {

  bool target_is_dir;
  IOOptions options;
  IODebugContext context;
  IOStatus status = target()->IsDirectory(dir, options, &target_is_dir, &context);
  // TODO if (!status.ok()) return status;
  if (!target_is_dir) {
    // TODO Return error
  }

  auto string_starts_with = [](std::string const& string,
                               std::string const& needle) {
    return string.rfind(needle, 0) == 0;
  };

  for (auto const& it : files_) {
    std::filesystem::path file_path(it.first);
    assert(file_path.has_filename());
    // if (!file_path.has_filename()) continue; // <- This should be impossible?

    std::string file_dir =
      file_path.parent_path();

    if (string_starts_with(file_dir, dir)) {
      if (include_grandchildren ||
          file_dir.length() == dir.length()) {
        result.push_back(file_path.filename().string());
      }
    }
  }
}
```

MaisenbacherD: This would not cover the following case:
Let's assume we call `GetZenFSChildrenNoLock("/path/exampledir", true, result_vector)` where `/path/exampledir` is a directory in the aux directory and `files_` contains a ZenFS file called `/path/exampledir2/examplefile`. 
The check if `/path/exampledir` is a directory would pass and `if (string_starts_with("/path/exampledir2", dir))` would also pass which means we would falsely consider `/path/exampledir2/examplefile` as a child of `/path/exampledir/`

metaspace: But then we could check if character after the prefix we search for is a separator. But maybe that is not better than what we have.

MaisenbacherD: Yep I think this is essentially the same check.

fs/fs_zenfs.cc
```diff
@@ -794,6 +794,81 @@ IOStatus ZenFS::GetFileSize(const std::string& f, const IOOptions& options,
...
+  GetZenFSChildrenNoLock(source_path, true, &children);
+
+  for (const auto& child : children) {
+    std::string source_child_path =
+        (std::filesystem::path(source_path) / std::filesystem::path(child))
+            .string();
+    std::string dest_child_path =
+        (std::filesystem::path(dest_path) / std::filesystem::path(child))
+            .string();
+    s = RenameFileNoLock(source_child_path, dest_child_path, options, dbg);
```
aravind-wdc, line 906: This code is repeated in RollbackAuxDirRenameNoLock() and here. One suggestion is can we simplify it by moving it to a separate function ?
pseudo code:
```
IOStatus UpdatePathAndRename(std::string src, std::string dest, std::vector<std::string> filelist, IOOptions& options, IODebugContext& dbg ) {
IOStatus s;
  for (const auto& child : filelist) {
    std::string source_child_path =
        (std::filesystem::path(src) / std::filesystem::path(child))
            .string();
    std::string dest_child_path =
        (std::filesystem::path(dest) / std::filesystem::path(child))
            .string();
    s = RenameFileNoLock(source_child_path, dest_child_path, options, dbg);
    if (!s.ok()) return s;
  }
  return s;
}

```

Remaining code can be same...

aravind-wdc (commented):
Just one comment/suggestion.
Looks good otherwise.

fs/fs_zenfs.cc
MaisenbacherD: This does not seem to simplify the code because of  `  renamed_children.push_back(child);` inside this for loop.

metaspace: I think @aravind-wdc suggests something like this:
```c++
IOStatus ZenFS::RenameChildren(std::string const& src_dir, std::string const& dest_dir,
            std::string const& child, const IOOptions& options,
            IODebugContext* dbg) {
  std::string source_rollback_child_path =
      (std::filesystem::path(src_dir) / std::filesystem::path(child)).string();
  std::string dest_rollback_child_path =
      (std::filesystem::path(dest_dir) / std::filesystem::path(child)).string();
  return RenameFileNoLock(source_rollback_child_path, dest_rollback_child_path,
                       options, dbg);
}

/* Must hold files_mtx_ */
IOStatus ZenFS::RollbackAuxDirRenameNoLock(
    const std::string& source_path, const std::string& dest_path,
    const std::vector<std::string>& renamed_children, const IOOptions& options,
    IODebugContext* dbg) {
  IOStatus s;

  for (const auto& rollback_child : renamed_children) {
    s = RenameChildren(dest_path, source_path, rollback_child, options, dbg);
    if (!s.ok()) {
      return IOStatus::Corruption(
          "RollbackAuxDirRenameNoLock: Failed to roll back directory rename");
    }
  }

  s = target()->RenameFile(ToAuxPath(dest_path), ToAuxPath(source_path),
                           options, dbg);
  if (!s.ok()) {
    return IOStatus::Corruption(
        "RollbackAuxDirRenameNoLock: Failed to roll back auxiliary path "
        "renaming");
  }

  return s;
}

/* Must hold files_mtx_ */
IOStatus ZenFS::RenameAuxPathNoLock(const std::string& source_path,
                                    const std::string& dest_path,
                                    const IOOptions& options,
                                    IODebugContext* dbg) {
  IOStatus s;
  std::vector<std::string> children;
  std::vector<std::string> renamed_children;

  s = target()->RenameFile(ToAuxPath(source_path), ToAuxPath(dest_path),
                           options, dbg);
  if (!s.ok()) {
    return s;
  }

  GetZenFSChildrenNoLock(source_path, true, &children);

  for (const auto& child : children) {
    s = RenameChildren(source_path, dest_path, child, options, dbg);
    if (!s.ok()) {
      IOStatus failed_rename = s;
      s = RollbackAuxDirRenameNoLock(source_path, dest_path, renamed_children,
                                     options, dbg);
      if (!s.ok()) {
        return s;
      }
      return failed_rename;
    }
    renamed_children.push_back(child);
  }

  return s;
}
```

MaisenbacherD: Sure that makes sense!

MaisenbacherD:
Addressed the suggestion of @aravind-wdc

aravind-wdc (approved):

metaspace (approved):
:+1:

yhr (approved):

PR #163, opened 2022-03-03
Fix file locking so functions can hold `files_mtx_` lock across file operations

aravind-wdc:


yhr:
Looks allright to me, but please squash the deletefile changes into a single commit

metaspace (changes requested):
I have one comment about the locking of `DeleteFile`, otherwise looks good.

I would prefer splitting the patches and merging across some of them so that we get one patch for each refactored method. If you want, I can split/merge the patches when review is done.

One thing we did not discuss is that we could probably avoid all these locking/non-locking functions if we change `files_mtx_` to a recursive lock. The problem in the first place is that we cannot grab the lock more than once, or we will deadlock. I'm not really opinionated on the topic, what do you guys think?

fs/fs_zenfs.cc
```diff
@@ -730,9 +710,8 @@ IOStatus ZenFS::OpenWritableFile(const std::string& fname,
   return s;
 }
 
-IOStatus ZenFS::DeleteFileNoLock(const std::string& fname,
-                                 const IOOptions& options,
-                                 IODebugContext* dbg) {
+IOStatus ZenFS::DeleteFile(const std::string& fname, const IOOptions& options,
+                           IODebugContext* dbg) {
```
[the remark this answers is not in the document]
aravind-wdc: Yes, I was actually fixing this now...  It needs to hold it across the function.

aravind-wdc:
> Looks allright to me, but please squash the deletefile changes into a single commit

Just wanted to ensure the new changes were reviewed so used a different commit. Like @metaspace suggested will split the commit into api specific commits, once review is done.

fs/fs_zenfs.cc
```diff
@@ -687,7 +667,7 @@ IOStatus ZenFS::GetChildren(const std::string& dir, const IOOptions& options,
 IOStatus ZenFS::OpenWritableFile(const std::string& fname,
                                  const FileOptions& file_opts,
                                  std::unique_ptr<FSWritableFile>* result,
-                                 IODebugContext* /*dbg*/, bool reopen) {
+                                 IODebugContext* dbg, bool reopen) {
```
metaspace, line 670: There is a concurrency bug here. Same kind as the one we are fixing with this patch set. If two threads call `OpenWritableFile()` with the same `fname` we are going to have a race on insertion into `files_`, and we end up with two `ZoneFile` objects with same `fname` but different `file_id_` and only one of them in `files_`. We need to lock across all of the method. Perhaps we can fix this in this PR as well?

aravind-wdc: Yes. I have force pushed the commit which holds the lock across the `OpenwriteableFile()` function.
This required some more optimization to DeleteFileNoLock() and also another boolean `resetIOZones` so that we do not call `ResetUnusedIOZones()` on all successful calls of `OpenWriteableFilename()`

It is part of this commit for now, if there are no more review comments, I will split the commits into smaller, area specific commits.

fs/fs_zenfs.cc
```diff
@@ -487,17 +487,21 @@ std::shared_ptr<ZoneFile> ZenFS::GetFile(std::string fname) {
 }
 
 /* Must hold files_mtx_ */
-IOStatus ZenFS::DeleteZoneFileNoLock(std::string fname) {
+IOStatus ZenFS::DeleteFileNoLock(std::string fname, const IOOptions& options,
```
indraneel-m, line 490: Have general concerns about the naming convention xyzNoLock().

Just noticed in the code that there are functions which follow exactly the same template (Lock to be held before calling the function) but they are differently named as xyzLocked() - examples are RollMetaZoneLocked(), WriteSnapshotLocked().

Then there are functions which again have similar template but have neither Locked nor NoLock suffix - example GetFileInternal().

Going back to the comment that I made earlier today in the Teams chat - this naming scheme is not very intuitive - we are naming functionality with similar pattern differently at different times. Let's conclude on a common scheme that's explains this best and document it for future use and clarity. @aravind-wdc , @MaisenbacherD , @metaspace , @yhr

metaspace: We should definitely settle on a scheme.

Of the three options, I think that  `xyzNoLock()` is the best one. It communicates that the operation will be performed without taking any locks. I also like `xyzLocked()` but it could be misinterpreted to mean that the operation will be performed locked in the sense that the method will take any locks it needs. `xyzInternal()` is not good, it says nothing about lock invariants.

aravind-wdc: @indraneel-m 
The `xyzLocked()`  naming conventions probably needs some of its own refactoring, which we can take up next.

yhr: Let's go with the NoLock naming scheme, and follow up with a PR aligning the rest of the code with this

metaspace:
Closed ?

fs/fs_zenfs.cc
```diff
@@ -477,40 +482,39 @@ std::shared_ptr<ZoneFile> ZenFS::GetFileInternal(std::string fname) {
...
+      /* Failed to persist the delete, return to a consistent state */
+      files_.insert(std::make_pair(fname.c_str(), zoneFile));
     } else {
-      s = IOStatus::NotFound("ZenFS::DeleteFile(): File not found");
+      /* Mark up the file as deleted so it won't be migrated by GC */
+      zoneFile->SetDeleted();
+      zoneFile.reset();
     }
+  } else {
+    s = target()->DeleteFile(ToAuxPath(fname), options, dbg);
```
metaspace, line 515: I don't think that we need to hold `files_mtx_` across deletion on aux fs. I does not break anything, but not strictly needed.

metaspace (approved):
I think you killed `IsDirectoryNoLock()`, which is required for the snapshot feature. I guess @MaisenbacherD can add it later.

fs/fs_zenfs.cc
indraneel-m: Sounds good..

aravind-wdc:
> I think you killed `IsDirectoryNoLock()`, which is required for the snapshot feature. I guess @MaisenbacherD can add it later.

Yes, I did not add it, I have the patch ready though which is applicable on top of this, I was thinking that it can go with couple of other patches in PR 160.

fs/fs_zenfs.cc
aravind-wdc: Yeah, I thought about this, if we specifically want to keep it out of the files_mtx_ we would need to introduce one more function and more plumbing is involved, so this seemed better than the alternative.
A recursive locking mechanism can be helpful to reduce these extra plumbing.

aravind-wdc:
Fixed the clang error.

yhr (approved):

yhr:
rebased and pushed manually

PR #167, opened 2022-03-07
Use `pkg-config` for specifying dependencies.

metaspace:
Use `pkg-config` to specify dependencies for ZenFS plugin. Required support was added to RocksDB a while back.


zenfs.mk
```diff
@@ -1,10 +1,12 @@
 zenfs_SOURCES = fs/fs_zenfs.cc fs/zbd_zenfs.cc fs/io_zenfs.cc
 zenfs_HEADERS = fs/fs_zenfs.h fs/zbd_zenfs.h fs/io_zenfs.h fs/version.h fs/metrics.h fs/snapshot.h
-zenfs_LDFLAGS = -lzbd -u zenfs_filesystem_reg
+zenfs_LDFLAGS = -u zenfs_filesystem_reg
```
aravind-wdc, line 3: Do we not need to link the zbd library anymore ?

metaspace: It's handled in line 12 by pkg-config.

aravind-wdc: ok

aravind-wdc (approved):

yhr (approved):

Issue #84, opened 2021-11-16
Enable logging for libzbd in debug builds


metaspace:
We can enable more logging in libzbd (printfs to stdout/sterr) when setting up a zbd inside zenfs.

indraneel-m:
@metaspace , can you elaborate a bit here? The ticket is for adding additional logging in libzbd or improved logging of the responses returned by libzbd to zenfs?

metaspace:
`libzbd` has internal logging infrastructure. It is disabled by default. The user of the library needs to call a setup function to enable logging. We had bugs earlier that would have been made visible by enabling error level logging in libzbd, so it is probably a good idea to do so.

yhr:
closed as it turned out impractical, see #171

PR #171, opened 2022-03-09
fs: Improve debugability by enabling libzbd error logging in zenfs.

indraneel-m:
Addresses issue #84 and improves debugability by enabling libzbd error logging in zenfs. Currently there's no debug build, hence the error logging level is set to ZBD_LOG_ERROR. Can be changed to ZBD_LOG_DEBUG if needed.

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -237,6 +237,10 @@ IOStatus ZonedBlockDevice::Open(bool readonly, bool exclusive) {
   if (!readonly && !exclusive)
     return IOStatus::InvalidArgument("Write opens must be exclusive");
 
+  // Enable error logging in libzbd before using libzbd apis.
+  // Change to ZBD_LOG_DEBUG for more verbose logging.
+  zbd_set_log_level(ZBD_LOG_ERROR);
+
   /* The non-direct file descriptor acts as an exclusive-use semaphore */
   if (exclusive) {
     read_f_ = zbd_open(filename_.c_str(), O_RDONLY | O_EXCL, &info);


indraneel-m:
Still not able to find a way to initialise the libzbd per thread variable for all threads. Need to come to a conclusion on how to proceed on this.

Tried the following code which initialises the variable for only 1 thread that calls the Open method.

class LibZbdSetLogLevel {
    public:
    LibZbdSetLogLevel() {
        zbd_set_log_level(ZBD_LOG_DEBUG);
#if (WITH_TESTS == OFF)
        zbd_set_log_level(ZBD_LOG_ERROR);
#endif
    }
};

IOStatus ZonedBlockDevice::Open(bool readonly, bool exclusive) {
  static thread_local LibZbdSetLogLevel zenfs_libzbd_log_level_init;
  ...
  ...
}

metaspace:
> Tried the following code which initialises the variable for only 1 thread that calls the Open method.

This will only initialize the variable for threads that call the `Open()` function. Did you try making it a global variable? I think that should work.

indraneel-m:
> > Tried the following code which initialises the variable for only 1 thread that calls the Open method.
> 
> This will only initialize the variable for threads that call the `Open()` function. Did you try making it a global variable? I think that should work.

If you make this a global, it's not initialised even for a single thread.

metaspace:
> If you make this a global, it's not initialised even for a single thread.

According to the [c++17 standard](https://www.cs.rit.edu/usr/local/pub/swm/n4660.pdf) section 6.7.2 basic.stc.thread, initialization is allowed to be lazy. It just has to occur before a thread uses the variable. So we have to make sure that all threads that use libzbd access the global. But at that point, we might as well just call `zbd_set_log_level()`.

Not sure what a good approach is here. What do people do in general logging frameworks?

yhr:
closing this as it turned out that the per-thread loglevel setting is unpractical

PR #172, opened 2022-03-09
test: Example integration test based on RocksDB gtest infrastructure

metaspace:
This patch adds a simple integration test based on RocksDB gtest infrastructure.
To be functional, the patch requires changes in rocksdb. The patch for rocksdb
is currenlty on top of the 6.29.3 branch and can be found [here](https://github.com/metaspace/rocksdb/tree/6.29.3-plugin-tests).

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -236,9 +236,9 @@ IOStatus ZenMetaLog::ReadRecord(Slice* record, std::string* scratch) {
   return IOStatus::OK();
 }
 
-ZenFS::ZenFS(ZonedBlockDevice* zbd, std::shared_ptr<FileSystem> aux_fs,
-             std::shared_ptr<Logger> logger)
-    : FileSystemWrapper(aux_fs), zbd_(zbd), logger_(logger) {
+ZenFS::ZenFS(std::unique_ptr<ZonedBlockDevice> zbd,
+             std::shared_ptr<FileSystem> aux_fs, std::shared_ptr<Logger> logger)
+    : FileSystemWrapper(aux_fs), zbd_(zbd.release()), logger_(logger) {
   Info(logger_, "ZenFS initializing");
   Info(logger_, "ZenFS parameters: block device: %s, aux filesystem: %s",
        zbd_->GetFilename().c_str(), target()->Name());
@@ -1396,13 +1396,14 @@ Status ZenFS::Mount(bool readonly) {
   return Status::OK();
 }
 
-Status ZenFS::MkFS(std::string aux_fs_path, uint32_t finish_threshold) {
+Status ZenFS::MkFS(std::filesystem::path const& aux_fs_path,
+                   uint32_t finish_threshold) {
   std::vector<Zone*> metazones = zbd_->GetMetaZones();
   std::unique_ptr<ZenMetaLog> log;
   Zone* meta_zone = nullptr;
   IOStatus s;
 
-  if (aux_fs_path.length() > 255) {
+  if (aux_fs_path.string().length() > 255) {
     return Status::InvalidArgument(
         "Aux filesystem path must be less than 256 bytes\n");
   }
@@ -1506,7 +1507,8 @@ Status NewZenFS(FileSystem** fs, const std::string& bdevname,
   }
 #endif
 
-  ZonedBlockDevice* zbd = new ZonedBlockDevice(bdevname, logger, metrics);
+  std::unique_ptr<ZonedBlockDevice> zbd{
+      new ZonedBlockDevice(bdevname, logger, metrics)};
   IOStatus zbd_status = zbd->Open(false, true);
   if (!zbd_status.ok()) {
     Error(logger, "mkfs: Failed to open zoned block device: %s",
@@ -1514,7 +1516,7 @@ Status NewZenFS(FileSystem** fs, const std::string& bdevname,
     return Status::IOError(zbd_status.ToString());
   }
 
-  ZenFS* zenFS = new ZenFS(zbd, FileSystem::Default(), logger);
+  ZenFS* zenFS = new ZenFS(std::move(zbd), FileSystem::Default(), logger);
   s = zenFS->Mount(false);
   if (!s.ok()) {
     delete zenFS;
--- a/fs/util.cc
+++ b/fs/util.cc
@@ -1,0 +1,153 @@
+
+#include "util.h"
+
+#include <dirent.h>
+#include <libzbd/zbd.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <cstdint>
+#include <filesystem>
+#include <memory>
+#include <system_error>
+
+#include "fs_zenfs.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+std::unique_ptr<ZonedBlockDevice> zbd_open(std::string const &zbd_path,
+                                           bool readonly, bool exclusive) {
+  std::unique_ptr<ZonedBlockDevice> zbd{
+      new ZonedBlockDevice(zbd_path, nullptr)};
+  IOStatus open_status = zbd->Open(readonly, exclusive);
+
+  if (!open_status.ok()) {
+    fprintf(stderr, "Failed to open zoned block device: %s, error: %s\n",
+            zbd_path.c_str(), open_status.ToString().c_str());
+    zbd.reset();
+  }
+
+  return zbd;
+}
+
+// Here we pass 'zbd' by non-const reference to be able to pass its ownership
+// to 'zenFS'
+Status zenfs_mount(std::unique_ptr<ZonedBlockDevice> &zbd,
+                   std::unique_ptr<ZenFS> *zenFS, bool readonly) {
+  Status s;
+
+  std::unique_ptr<ZenFS> localZenFS{
+      new ZenFS(zbd.release(), FileSystem::Default(), nullptr)};
+  s = localZenFS->Mount(readonly);
+  if (!s.ok()) {
+    localZenFS.reset();
+  }
+  *zenFS = std::move(localZenFS);
+
+  return s;
+}
+
+static int is_dir(const char *path) {
+  struct stat st;
+  if (stat(path, &st) != 0) {
+    fprintf(stderr, "Failed to stat %s\n", path);
+    return 1;
+  }
+  return S_ISDIR(st.st_mode);
+}
+
+// Create or check pre-existing aux directory and fail if it is
+// inaccessible by current user and if it has previous data
+static int create_aux_dir(const char *path) {
+  struct dirent *dent;
+  size_t nfiles = 0;
+  int ret = 0;
+
+  errno = 0;
+  ret = mkdir(path, 0750);
+  if (ret < 0 && EEXIST != errno) {
+    fprintf(stderr, "Failed to create aux directory %s: %s\n", path,
+            strerror(errno));
+    return 1;
+  }
+  // The aux_path is now available, check if it is a directory infact
+  // and is empty and the user has access permission
+
+  if (!is_dir(path)) {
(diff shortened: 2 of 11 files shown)


fs/fs_zenfs_test.cc
```diff
@@ -0,0 +1,49 @@
...
+ public:
+  std::shared_ptr<FileSystem> fs_;
+
+  FsZenFSTest() {}
+
+ protected:
+  void SetUp() override {
+    ConfigOptions config_options;
+    config_options.ignore_unsupported_options = false;
+    char* fs_uri_env = std::getenv("DEVICE");
```
aravind-wdc, line 22: This can be "zns_device" or just "device"

fs/fs_zenfs_test.cc
```diff
@@ -0,0 +1,49 @@
...
+    ASSERT_TRUE(fs_uri_env != nullptr) << "DEVICE environment variable not set";
+    std::string fs_uri = "zenfs://dev:" + std::string(fs_uri_env);
```
aravind-wdc, line 24: Making this "fs_uri_env" would be better.

fs/fs_zenfs_test.cc
```diff
@@ -0,0 +1,49 @@
...
+  void TearDown() override {}
+};
+
+TEST_F(FsZenFSTest, Example) {
+  IOStatus status;
+
+  ASSERT_TRUE(1);
+
+  IOOptions options;
+  status = fs_->CreateDir("test", options, nullptr);
```
aravind-wdc, line 38: This creates the directory in aux-path. 
As an example we can add something happening within ZenFS. Like create a file maybe.

aravind-wdc (commented):
Looks good in general, apart from a couple of nitpicks.
Adding a small Readme or a Guide on how to add more tests or run the tests, will help a lot.

fs/fs_zenfs_test.cc
metaspace: Not sure what you mean. The `DEVICE` environment variable? It should be a zoned storage device. I'm using `null_blk` for testing.

fs/fs_zenfs_test.cc
metaspace: Not sure what you mean. `fs_uri_env` identifier is used for the environment variable. It's not really an URI, so maybe rename it to device_env would make sense.

fs/fs_zenfs_test.cc
metaspace: > This creates the directory in aux-path.

Yes, that was the intention.

> As an example we can add something happening within ZenFS. Like create a file maybe.

For sure. This is just a placeholder to show that the infrastructure is working and how to add new test cases. I imagine that we want a whole bunch of tests to cover as much of ZenFS as possible.

Perhaps we should make a habit of adding a test each time we find a bug so that the test triggers said bug. Sort of TDD for bug fixing.

metaspace:
> Adding a small Readme or a Guide on how to add more tests or run the tests, will help a lot.

I'll update the ZenFS docs with instructions on how to execute and pointers to `gtest` docs.

metaspace:
Now that I think of it, I think that it would be better if the test suite create the ZenFS filesystem in `SetUp()` rather than relying on it being present on the device. Will fix that.

zenfs.mk
```diff
@@ -1,6 +1,7 @@
 zenfs_SOURCES = fs/fs_zenfs.cc fs/zbd_zenfs.cc fs/io_zenfs.cc
 zenfs_HEADERS = fs/fs_zenfs.h fs/zbd_zenfs.h fs/io_zenfs.h fs/version.h fs/metrics.h fs/snapshot.h
 zenfs_LDFLAGS = -u zenfs_filesystem_reg
+zenfs_TEST_MAIN_SOURCES = fs/fs_zenfs_test.cc
```
yhr, line 4: I think we should introduce a test/ directory with the tests in stead

yhr (commented):
Looks good over all i think. The example is illustrative, but maybe it would be good to add a couple of "real" tests to start off with?

fs/fs_zenfs_test.cc
metaspace: Changed the naming of the variables.

fs/fs_zenfs_test.cc
metaspace: Changed names of these locals

zenfs.mk
metaspace: Moved tests to subdir

MaisenbacherD (commented):
For commit c5f4e30b3f1136eb62d31be09cae496e51d437b2: `.github/workflows/mtr-test.yaml` would also need to bump percona-server to `Percona-Server-8.0.28-19`

fs/fs_zenfs.cc
```diff
@@ -1254,13 +1255,14 @@ Status ZenFS::Mount(bool readonly) {
   return Status::OK();
 }
 
-Status ZenFS::MkFS(std::string aux_fs_path, uint32_t finish_threshold) {
+Status ZenFS::MkFS(std::filesystem::path const& aux_fs_path,
```
MaisenbacherD, line 1258: The line `Superblock super(zbd_, aux_fs_path, finish_threshold);` should be converted to `Superblock super(zbd_, aux_fs_path.string(), finish_threshold);`

fs/util.cc
```diff
@@ -16,143 +13,151 @@
...
-    fprintf(stderr, "Failed to create aux directory %s: %s\n", path,
-            strerror(errno));
-    return 1;
+static Status create_aux_dir(std::filesystem::path const &path) {
+  namespace fs = std::filesystem;
+  std::error_code ec;
+
+  bool aux_exists = fs::exists(path, ec);
+  if (ec) {
+    return Status::Aborted("Aux directory exists" + ec.message());
```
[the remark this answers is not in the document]
metaspace: I think i did this because i got nonzero `ec` when the path did not exist. Makes no sense though. I will check again.

util/zenfs.cc
```diff
@@ -154,20 +160,22 @@ int zenfs_tool_list() {
...
+  std::unique_ptr<ZonedBlockDevice> zbd;
   std::unique_ptr<ZenFS> zenFS;
-  s = zenfs_mount(zbd, &zenFS, true);
+
+  s = zenfs_zbd_open(FLAGS_zbd, true, false, zbd);
+  if (!s.ok() || !zbd) return 1;
+
+  uint64_t used = zbd->GetUsedSpace();
+  uint64_t free = zbd->GetFreeSpace();
+  uint64_t reclaimable = zbd->GetReclaimableSpace();
```
MaisenbacherD, line 170: Is there a reason why `used`, `free` and `reclaimable` are now assigned before checking for a successful mount?

util/zenfs.cc
```diff
@@ -578,21 +600,23 @@ int zenfs_tool_restore() {
...
-
+  std::unique_ptr<ZonedBlockDevice> zbd;
   std::unique_ptr<ZenFS> zenFS;
-  s = zenfs_mount(zbd, &zenFS, true);
+  std::ostream &json_stream = std::cout;
+
+  s = zenfs_zbd_open(FLAGS_zbd, true, false, zbd);
+  if (!s.ok() || !zbd) return 1;
+
+  json_stream << "{\"zones\":";
```
MaisenbacherD, line 609: Why did the initial setup of the `json_stream` went before the check of a successful mount?

util/zenfs.cc
```diff
@@ -374,7 +382,12 @@ int zenfs_tool_backup() {
   IOStatus io_status;
   IOOptions opts;
   IODebugContext dbg;
-  std::unique_ptr<ZonedBlockDevice> zbd = zbd_open(FLAGS_zbd, true, true);
+  std::unique_ptr<ZonedBlockDevice> zbd;
+
+  status = zenfs_zbd_open(FLAGS_zbd, true, true, zbd);
```
MaisenbacherD, line 386: After calling `zenfs_zbd_open` sometimes only the status is checked, and sometimes the `zbd` pointer is additionally checked to not be `null`. The check should always include `!zbd` or not?

MaisenbacherD (commented):
Had some comments :)

Also commit fa829de0f39739bb08ab9d6cca8c113e59562c0c is referring probably to the wrong issue.

util/zenfs.cc
metaspace: Yes, once you pass the zbd too ZenFS, you cannot call it. Holding on to the raw pointer of a unique_ptr defeats the purpose.

util/zenfs.cc
metaspace: Same as above. We loose the zbd handle when mounting.

util/zenfs.cc
metaspace: It is proper defensive programming to check both values, I guess. But I think both are checked? Some locations zbd is checked further down in code that is not touched by this patch.

fs/util.cc
```diff
@@ -16,143 +13,151 @@
...
+      std::unique_ptr<ZonedBlockDevice>{new ZonedBlockDevice(path, nullptr)};
+  IOStatus status = out_zbd->Open(readonly, exclusive);
 
-  if (!open_status.ok()) {
+  if (!status.ok()) {
     fprintf(stderr, "Failed to open zoned block device: %s, error: %s\n",
-            zbd_path.c_str(), open_status.ToString().c_str());
-    zbd.reset();
+            path.c_str(), status.ToString().c_str());
+    out_zbd.reset();
```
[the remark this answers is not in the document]
aravind-wdc: Ok, I see that.
But we reset it only when status is not ok.
And in calling context, if status is not ok, we anyway return which would destruct the zbd object, right ?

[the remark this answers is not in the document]
metaspace: We reset the smart pointer on the error path in order to not return a partially initialized object to the caller. You are right, the caller should never access the object if the returned status is not OK. The object would be destroyed when the unique_ptr goes out of scope in the calling context, but only if zbd is a reference to a variable in the calling context. It might be from higher in the call stack or even on the heap, we cannot know. By calling reset, we ensure object destruction before function return, and we ensure that the unique_ptr will be empty, wherever it is allocated. 

We could state in the API contract that the value of the zbd argument is undefined and should not be used in case status is not OK. But then we would rely on the caller doing something that we could just assert before returning. Resetting it before return on the error path is a defensive programming strategy that makes it more probable that the API is used the way we intend.

Also, in case the caller wants to retry the call, it should not be the task of the caller to destroy a partially initialized object from the zbd argument. Better to keep the invariant that if an object is returned, it is fully initialized.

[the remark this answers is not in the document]
metaspace: If `zbd` was a regular pointer pointer, we would also set that to `nullptr`.

fs/fs_zenfs.cc
metaspace: I think passing the path and using implicit conversion is fine. There is a conversion function on `path` to `string_type`.
 
* [conversion operator](https://en.cppreference.com/w/cpp/language/cast_operator)
* [implicit conversion](https://en.cppreference.com/w/cpp/language/implicit_conversion)
* [conversion operator on path](https://en.cppreference.com/w/cpp/filesystem/path/native)

fs/util.cc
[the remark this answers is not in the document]
metaspace: Changed the error message.

metaspace:
> For commit [c5f4e30](https://github.com/westerndigitalcorporation/zenfs/commit/c5f4e30b3f1136eb62d31be09cae496e51d437b2): `.github/workflows/mtr-test.yaml` would also need to bump percona-server to `Percona-Server-8.0.28-19`

Bumped percona server to 8.0.28-19.

metaspace:
Rebased on master

metaspace:
Seems github is broken. I will close this PR and open a new one.

Issue #170, opened 2022-03-09
"zenfs list" command does not list files when path starts with ./" 


aravind-wdc:
```
$ sudo ./bin/zenfs list --zbd=nvme0n1 --path=./.rocksdb                                                           
           0    Mar 09 2022 13:19:16            LOCK
       56525    Mar 09 2022 13:19:19            LOG.old.1646812163053548
       70558    Mar 09 2022 13:20:31            LOG
```
There are sst files in ZenFS written by MyRocks, but it is not getting listed.

aravind-wdc:
This is due to 
https://github.com/westerndigitalcorporation/zenfs/blob/master/util/zenfs.cc#L68

MaisenbacherD:
Related issue: https://github.com/westerndigitalcorporation/zenfs/issues/164

yhr:
We can use c++ 17 path normalization to convert all filenames/paths coming in, to avoid problems comparing file name formats internally. 

See
https://en.cppreference.com/w/cpp/filesystem/path/lexically_normal

aravind-wdc:
Resolved by #176

PR #176, opened 2022-03-25
fs: format all input paths into ZenFs lexically

aravind-wdc:
It formats input paths lexically, to handle scenarios like
relative paths, superfluous slashes.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -15,7 +15,6 @@
 #include <string.h>
 #include <unistd.h>
 
-#include <filesystem>
 #include <sstream>
 #include <utility>
 #include <vector>
@@ -272,6 +271,11 @@ IOStatus ZenFS::Repair() {
   return IOStatus::OK();
 }
 
+std::string ZenFS::FormatPathLexically(std::filesystem::path filepath) {
+  std::filesystem::path ret = "/" / filepath.lexically_normal();
+  return ret.string();
+}
+
 void ZenFS::LogFiles() {
   std::map<std::string, std::shared_ptr<ZoneFile>>::iterator it;
   uint64_t total_size = 0;
@@ -474,6 +478,7 @@ IOStatus ZenFS::SyncFileMetadata(ZoneFile* zoneFile, bool replace) {
 /* Must hold files_mtx_ */
 std::shared_ptr<ZoneFile> ZenFS::GetFileNoLock(std::string fname) {
   std::shared_ptr<ZoneFile> zoneFile(nullptr);
+  fname = FormatPathLexically(fname);
   if (files_.find(fname) != files_.end()) {
     zoneFile = files_[fname];
   }
@@ -493,6 +498,7 @@ IOStatus ZenFS::DeleteFileNoLock(std::string fname, const IOOptions& options,
   std::shared_ptr<ZoneFile> zoneFile(nullptr);
   IOStatus s;
 
+  fname = FormatPathLexically(fname);
   zoneFile = GetFileNoLock(fname);
   if (zoneFile != nullptr) {
     std::string record;
@@ -522,10 +528,11 @@ IOStatus ZenFS::DeleteFileNoLock(std::string fname, const IOOptions& options,
   return s;
 }
 
-IOStatus ZenFS::NewSequentialFile(const std::string& fname,
+IOStatus ZenFS::NewSequentialFile(const std::string& filename,
                                   const FileOptions& file_opts,
                                   std::unique_ptr<FSSequentialFile>* result,
                                   IODebugContext* dbg) {
+  std::string fname = FormatPathLexically(filename);
   std::shared_ptr<ZoneFile> zoneFile = GetFile(fname);
 
   Debug(logger_, "New sequential file: %s direct: %d\n", fname.c_str(),
@@ -540,10 +547,11 @@ IOStatus ZenFS::NewSequentialFile(const std::string& fname,
   return IOStatus::OK();
 }
 
-IOStatus ZenFS::NewRandomAccessFile(const std::string& fname,
+IOStatus ZenFS::NewRandomAccessFile(const std::string& filename,
                                     const FileOptions& file_opts,
                                     std::unique_ptr<FSRandomAccessFile>* result,
                                     IODebugContext* dbg) {
+  std::string fname = FormatPathLexically(filename);
   std::shared_ptr<ZoneFile> zoneFile = GetFile(fname);
 
   Debug(logger_, "New random access file: %s direct: %d\n", fname.c_str(),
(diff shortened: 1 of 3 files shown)


aravind-wdc:
This also fixes: https://github.com/westerndigitalcorporation/zenfs/issues/170
Tests passed:
  smoke-test,
  util-tests, 
  quick-performance,
  valgrind tests and backup/restore tests mentioned in WDCZEN-26, 
  percona-server all benchmark scripts,
  dbbench runs,
  manual tests.

fs/fs_zenfs.cc
```diff
@@ -272,6 +272,11 @@ IOStatus ZenFS::Repair() {
   return IOStatus::OK();
 }
 
+std::string ZenFS::FormatPathLexically(std::string filepath) {
+  std::filesystem::path f = filepath;
+  return f.lexically_normal().string();
+}
+
```
metaspace, line 279: We can take a `std::filesystem::path` as argument here, conversion from `std::string` is implicit. If we want to keep `std::string`, take the string as `const&` to avoid invoking copy constructor on function invocation.

fs/fs_zenfs.cc
```diff
@@ -522,10 +529,11 @@ IOStatus ZenFS::DeleteFileNoLock(std::string fname, const IOOptions& options,
   return s;
 }
 
-IOStatus ZenFS::NewSequentialFile(const std::string& fname,
+IOStatus ZenFS::NewSequentialFile(const std::string& filename,
                                   const FileOptions& file_opts,
                                   std::unique_ptr<FSSequentialFile>* result,
                                   IODebugContext* dbg) {
+  std::string fname = FormatPathLexically(filename);
```
MaisenbacherD, line 536: Maybe we could refactor `fname` to be something like `normalized_filename`.

metaspace (changes requested):
Looks good to me, just one comment.

MaisenbacherD (commented):
Should we introduce the path normalisation in each method that handles paths or file names as a defensive measure (eg. in `ZenFS::MigrateFileExtents`)?

fs/fs_zenfs.cc
```diff
@@ -272,6 +272,11 @@ IOStatus ZenFS::Repair() {
   return IOStatus::OK();
 }
 
+std::string ZenFS::FormatPathLexically(std::string filepath) {
+  std::filesystem::path f = filepath;
+  return f.lexically_normal().string();
```
yhr, line 277: Based on yura's feedback, they expect that "/test.txt" and "test.txt" paths should be equal. So I think the easiest here is to add a prefix slash to the paths, something like:

``` 
 std::string normal = f.lexically_normal().string();
 if (normal.size() == 0 || normal.at(0) != '/') {
   normal = "/" + normal;
 }
 return normal;

File: util/zenfs.cc [hunk 4]
```diff
@@ -53,27 +53,6 @@ bool zenfs_format_path(std::string &path, bool directory) {
   /* if the path is a directory, ensure that it ends with a slash */
   if (directory && path.back() != '/') path = path + "/";
 
-  /* relative paths are not supported */
```
yhr, line 56: After this change the function does not do much and will never fail, so I think we cshould just drop this function and open-code the slash additions in the two places it is used..

yhr (changes requested):

yhr:
> 



> Should we introduce the path normalisation in each method that handles paths or file names as a defensive measure (eg. in `ZenFS::MigrateFileExtents`)?

That should not be needed, as all entry points are covered

fs/fs_zenfs.cc
aravind-wdc: Ok.

fs/fs_zenfs.cc
aravind-wdc: Ok

fs/fs_zenfs.cc
aravind-wdc: I did that in the first iteration for a couple of APIs, but it needed changing the name across the function causing a lot of unnecessary code churn, for no extra benefit, so chose this way.

fs/fs_zenfs.cc
metaspace: @yhr we can unconditionally  add the `/` to the front before calling `lexically_normal()`, then we will always have just 1 leading slash.

Not sure what the check for size is about? Even if path is empty or path is ".", we still want a leading slash, right?

util/zenfs.cc
metaspace: I disagree, removing a function to duplicate a line of code does not seem right to me.

yhr: A one-liner add adding a slash if needed does not mandate a function i think, but it does not hurt. the function needs to be rewritten in any case.

fs/fs_zenfs.cc
yhr: Need to check the size before checking index 0 to avoid going out of bounds

util/zenfs.cc
aravind-wdc: This function in its current form was returning always true and i called from list, backup, restore paths, but it is not necessary from "zenfs backup" path. So I have removed that.
So in place checks of "/" is needed in only 2 places (list, restore).

I have moved this check to a helper function.

yhr (approved):
Looks alright to me, and I've done some testing with backup/restore/list and it looks like its all behaving like expected

fs/fs_zenfs.cc
metaspace: Always refactor if it cleans up the code :+1:

fs/fs_zenfs.cc
```diff
@@ -272,6 +271,13 @@ IOStatus ZenFS::Repair() {
   return IOStatus::OK();
 }
 
+std::string ZenFS::FormatPathLexically(std::filesystem::path filepath) {
+  std::string normal = filepath.lexically_normal().string();
+  if (normal.size() == 0 || normal.at(0) != '/') normal = "/" + normal;
```
metaspace, line 276:
```suggestion
  std::filesystem::path ret =   "/" / filepath.lexically_normal();
  return ret.string()
```

metaspace (changes requested):

fs/fs_zenfs.cc
yhr: And we need to check of the slash is present, as we don't want double-slashes if the user has already provided a slash in the file name (lexically_normal does not remove it)

metaspace: See my change suggestion below. Just add the slash before converting to string, then there is no issue. `std::filesystem::path` will remove double slashes.

fs/fs_zenfs.cc
yhr: this ensures a single prefixing slash?

MaisenbacherD (approved):

PR #177, opened 2022-03-28
Unit tests

metaspace:
This is a retry of #172 because the DCO check for some reason was not triggered

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,28 @@ This is done with the zenfs utility, using the mkfs command:
 ./plugin/zenfs/util/zenfs mkfs --zbd=<zoned block device> --aux_path=<path to store LOG and LOCK files>
 ```
 
+## Test
+
+ZenFS uses RocksDB gtest infrastructure for testing. See
+[gtest](https://google.github.io/googletest/) documentation for details on the
+test framework and how to add tests. ZenFS tests require a zoned block device to operate. Pass the name of the device in the `DEVICE` environment variable. You do not need to create a ZenFS filesystem on the device, the test will do that. *Note that all data on the device will be lost*.
+
+To run all RocksDB tests:
+
+```shell
+make DEBUG_LEVEL=1 ROCKSDB_PLUGINS=zenfs DEVICE=nullb0 check
+```
+See RocksDB `Makefile` for details on the `check` target.
+
+To build a single test:
+```shell
+make DEBUG_LEVEL=1 ROCKSDB_PLUGINS=zenfs fs_zenfs_test
+```
+
+To run the test, set `DEVICE` environment variable and execute the test:
+```shell
+DEVICE=null0 ./fs_zenfs_test
+```
 ## ZenFS on-disk file formats
 
 ZenFS Version 1.0.0 and earlier uses version 1 of the on-disk format.
--- a/fs/test/fs_zenfs_test.cc
+++ b/fs/test/fs_zenfs_test.cc
@@ -4,9 +4,9 @@
 
 #include <gtest/gtest.h>
 
-#include <cstdlib>
 #include <filesystem>
 #include <memory>
+#include <string>
 
 #include "../util.h"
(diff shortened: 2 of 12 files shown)


fs/test/fs_zenfs_test.cc
```diff
@@ -0,0 +1,83 @@
...
+  IOStatus status;
+
+  ASSERT_TRUE(1);
+
+  IOOptions options;
+  status = fs_->CreateDir("test", options, nullptr);
+  ASSERT_TRUE(status.ok()) << "Failed to create directory 'test'";
+
+  status = fs_->CreateDir("test", options, nullptr);
+  ASSERT_TRUE(!status.ok()) << "Directory 'test' could be created twice";
```
aravind-wdc, line 76: Just one comment / suggestion:
I assume ASSERT_TRUE() would assert and fail the test on a test case failure. If so, it is fine for cases like mkfs failure. But for cases like "directory creation failure" or "same directory created twice" which may not be critical, can we look at allowing the subsequent tests to continue after reporting the failed test case (in stead of failing the test suite by asserting) ?

fs/test/fs_zenfs_test.cc
```diff
@@ -0,0 +1,83 @@
...
+#include "rocksdb/env.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+class FsZenFSTest : public testing::Test {
+ private:
+  std::filesystem::path aux_path;
+
+  void uniqueue_path(std::filesystem::path& out_path) {
+    char* name = std::tmpnam(nullptr);
```
MaisenbacherD, line 22: Is there a specific reason to use `tmpnam`? When compiling the test case I am getting the following warning: 
```
warning: the use of `tmpnam' is dangerous, better use `mkstemp'`
```

MaisenbacherD (commented):
One comment on a warning that I am getting when compiling. Otherwise looks good to me!

fs/test/fs_zenfs_test.cc
MaisenbacherD: Probably we want to ignore that warning because we just want to return a unique path that is then created by ZenFS's mkfs.

fs/test/fs_zenfs_test.cc
MaisenbacherD: Looks like the test suite continues after failing on a test case. https://google.github.io/googletest/primer.html:
> 4. When tests fail, they should provide as much information about the problem as possible. googletest doesn’t stop at the first test failure. Instead, it only stops the current test and continues with the next.

README.md
```diff
@@ -67,6 +67,28 @@ This is done with the zenfs utility, using the mkfs command:
 ./plugin/zenfs/util/zenfs mkfs --zbd=<zoned block device> --aux_path=<path to store LOG and LOCK files>
 ```
 
+## Test
```
MaisenbacherD, line 70: Could you also update the `clang-format-11` calls in the `Contribution Guide` section to include the test files? :)

fs/test/fs_zenfs_test.cc
metaspace: @aravind-wdc `ASSERT_TRUE()` terminates the current test case (function). There are also methods that allow the current test case to proceed, these are prefixed with `EXPECT_`. See https://google.github.io/googletest/reference/assertions.html

fs/test/fs_zenfs_test.cc
metaspace: @MaisenbacherD  I am OK with the risk in this test case.

metaspace:
Added a test from @MaisenbacherD to this PR.

metaspace:
Rebased on master

yhr:
We'll need to rebase and extend the testing framework to make it useful. Let's do that when we add a significant amount of functionality. I've parked the code for now inhttps://github.com/westerndigitalcorporation/zenfs/tree/unit-tests

PR #180, opened 2022-04-06
Fix extent blocks compute error

hml1006:
Ex: extent_length is 4094, sparse head+data+pad, it needs 2 blocks, not 1


yhr:
@hml1006 : Nice catch, looks good to me! but could you signoff (git commit --signoff --amend) the patch and push that?

hml1006:
> @hml1006 : Nice catch, looks good to me! but could you signoff (git commit --signoff --amend) the patch and push that?

Done

yhr:
@hml1006 : merged this through #182 after squashing the patches and fixing up the commit message a bit. Thanks!

PR #183, opened 2022-04-11
fs: Fix `GetZenFSChildrenNoLock` to return relative children paths.

MaisenbacherD:
When calling `GetZenFSchildrenNoLock` with `include_grandchildren == true`,
the reported grandchilden need to contain the complete relative path
from the given `dir` on.


yhr (approved):

aravind-wdc (approved):

fs/fs_zenfs.cc
```diff
@@ -652,6 +652,11 @@ void ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
   std::string dir_with_terminating_seperator =
       path_as_string_with_separator_at_end(std::filesystem::path(dir));
 
+  auto relative_child_path =
+      [dir_with_terminating_seperator](std::string const& full_path) {
```
metaspace, line 656:
```suggestion
      [&dir_with_terminating_seperator](std::string const& full_path) {
```
No need to copy the path, it is probably fine to take a reference.

metaspace (changes requested):
Functionally fine, 1 change suggestion.

fs/fs_zenfs.cc
MaisenbacherD: Fixed :)

metaspace (approved):

PR #184, opened 2022-04-11
Garbage collection

yhr:
Add an option to enable garbage collection in zenfs.

The garbage collection, if enabled, kicks in under 20% capacity left and cleans out zones with 1% valid data in them to start off with, increasing to 40% at 99% capacity.

This scheme reduces space amplification down to as little as 5% (from ~10-15%) in my testing.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -106,6 +106,8 @@ void Superblock::GetReport(std::string* reportString) {
   reportString->append(std::to_string(nr_zones_));
   reportString->append("\nFinish Threshold [%]:\t\t");
   reportString->append(std::to_string(finish_treshold_));
+  reportString->append("\nGarbage Collection Enabled:\t");
+  reportString->append(std::to_string(!!(flags_ & FLAGS_ENABLE_GC)));
   reportString->append("\nAuxiliary FS Path:\t\t");
   reportString->append(aux_fs_path_);
   reportString->append("\nZenFS Version:\t\t\t");
@@ -1428,7 +1430,8 @@ Status ZenFS::Mount(bool readonly) {
   return Status::OK();
 }
 
-Status ZenFS::MkFS(std::string aux_fs_p, uint32_t finish_threshold) {
+Status ZenFS::MkFS(std::string aux_fs_p, uint32_t finish_threshold,
+                   bool enable_gc) {
   std::vector<Zone*> metazones = zbd_->GetMetaZones();
   std::unique_ptr<ZenMetaLog> log;
   Zone* meta_zone = nullptr;
@@ -1472,7 +1475,7 @@ Status ZenFS::MkFS(std::string aux_fs_p, uint32_t finish_threshold) {
 
   log.reset(new ZenMetaLog(zbd_, meta_zone));
 
-  Superblock super(zbd_, aux_fs_path, finish_threshold);
+  Superblock super(zbd_, aux_fs_path, finish_threshold, enable_gc);
   std::string super_string;
   super.EncodeTo(&super_string);
 
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <set>
 #include <sstream>
 #include <utility>
 #include <vector>
@@ -258,9 +259,65 @@ ZenFS::~ZenFS() {
   zbd_->LogZoneUsage();
   LogFiles();
 
+  if (gc_worker_) {
+    run_gc_worker_ = false;
+    gc_worker_->join();
+  }
+
   meta_log_.reset(nullptr);
   ClearFiles();
   delete zbd_;
+}
+
+void ZenFS::GCWorker() {
+  while (run_gc_worker_) {
+    usleep(1000 * 1000 * 10);
+
+    uint64_t non_free = zbd_->GetUsedSpace() + zbd_->GetReclaimableSpace();
+    uint64_t free = zbd_->GetFreeSpace();
+    uint64_t free_percent = (100 * free) / (free + non_free);
+    ZenFSSnapshot snapshot;
+    ZenFSSnapshotOptions options;
+
+    if (free_percent > GC_START_LEVEL) continue;
+
+    options.zone_ = 1;
+    options.zone_file_ = 1;
+    options.log_garbage_ = 1;
+
+    GetZenFSSnapshot(snapshot, options);
+
+    uint64_t threshold = (100 - GC_SLOPE * (GC_START_LEVEL - free_percent));
+    std::set<uint64_t> migrate_zones_start;
+    for (const auto& zone : snapshot.zones_) {
+      if (zone.capacity == 0) {
+        uint64_t garbage_percent_approx =
+            100 - 100 * zone.used_capacity / zone.max_capacity;
+        if (garbage_percent_approx > threshold &&
+            garbage_percent_approx < 100) {
+          migrate_zones_start.emplace(zone.start);
+        }
+      }
+    }
+
+    std::vector<ZoneExtentSnapshot*> migrate_exts;
+    for (auto& ext : snapshot.extents_) {
+      if (migrate_zones_start.find(ext.zone_start) !=
+          migrate_zones_start.end()) {
+        migrate_exts.push_back(&ext);
+      }
+    }
+
+    if (migrate_exts.size() > 0) {
+      IOStatus s;
+      Info(logger_, "Garbage collecting %d extents \n",
+           (int)migrate_exts.size());
+      s = MigrateExtents(migrate_exts);
+      if (!s.ok()) {
+        Error(logger_, "Garbage collection failed");
+      }
+    }
+  }
 }
 
 IOStatus ZenFS::Repair() {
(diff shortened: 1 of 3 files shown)


fs/fs_zenfs.cc
```diff
@@ -255,11 +256,65 @@ ZenFS::~ZenFS() {
...
+    options.log_garbage_ = 1;
+
+    GetZenFSSnapshot(snapshot, options);
+
+    uint64_t threshold = (100 - GC_SLOPE * (GC_START_LEVEL - free_percent));
+    std::set<uint64_t> migrate_zone_ids;
+    for (const auto& zone : snapshot.zones_) {
+      if (zone.capacity == 0) {
+        uint64_t garbage_percent =
+            100 - 100 * zone.used_capacity / zone.max_capacity;
```
MaisenbacherD, line 292: Maybe we should consider refactoring `zone.used_capacity` to something like `zone.valid_data_capacity`.

fs/fs_zenfs.cc
```diff
@@ -255,11 +256,65 @@ ZenFS::~ZenFS() {
...
+  while (run_gc_worker_) {
+    uint64_t non_free = zbd_->GetUsedSpace() + zbd_->GetReclaimableSpace();
+    uint64_t free = zbd_->GetFreeSpace();
+    uint64_t free_percent = (100 * free) / (free + non_free);
+    ZenFSSnapshot snapshot;
+    ZenFSSnapshotOptions options;
+
+    usleep(1000 * 1000 * 10);
+
+    if (free_percent > GC_START_LEVEL) continue;
```
MaisenbacherD, line 279: Maybe we should add another threshold that prevents reclaiming zones if there is not much reclaimable space.

fs/fs_zenfs.cc
```diff
@@ -255,11 +256,65 @@ ZenFS::~ZenFS() {
...
+
+    GetZenFSSnapshot(snapshot, options);
+
+    uint64_t threshold = (100 - GC_SLOPE * (GC_START_LEVEL - free_percent));
+    std::set<uint64_t> migrate_zone_ids;
+    for (const auto& zone : snapshot.zones_) {
+      if (zone.capacity == 0) {
+        uint64_t garbage_percent =
+            100 - 100 * zone.used_capacity / zone.max_capacity;
+        if (garbage_percent > threshold && garbage_percent < 100) {
```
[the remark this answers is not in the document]
metaspace: @MaisenbacherD  If `garbage_percent == 100`, no migration is required. The zone will be picked up by `ZonedBlockDevice::ResetUnusedIOZones()` at some point.

fs/fs_zenfs.cc
```diff
@@ -253,11 +256,65 @@ ZenFS::~ZenFS() {
...
+    options.zone_file_ = 1;
+    options.log_garbage_ = 1;
+
+    GetZenFSSnapshot(snapshot, options);
+
+    uint64_t threshold = (100 - GC_SLOPE * (GC_START_LEVEL - free_percent));
+    std::set<uint64_t> migrate_zone_ids;
+    for (const auto& zone : snapshot.zones_) {
+      if (zone.capacity == 0) {
+        uint64_t garbage_percent =
```
metaspace, line 291:
```suggestion
        uint64_t garbage_percent_approximate =
```

The snapshots are not consistent in the face of concurrency. It sort of bothers me, but I don't think it will break anything. Just be aware that the numbers might not add up.

metaspace (changes requested):
I don't think it is safe to call `MigrateExtents` concurrently like this. We would need to make `MigrateExtents` thread safe or stop the world while migrating.

fs/fs_zenfs.cc
```diff
@@ -255,11 +256,65 @@ ZenFS::~ZenFS() {
   zbd_->LogZoneUsage();
   LogFiles();
 
+  if (gc_worker_ != nullptr) {
+    run_gc_worker_ = false;
+    gc_worker_->join();
```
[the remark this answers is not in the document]
yhr: Yep

fs/fs_zenfs.cc
yhr: Let's think about naming and do that as a separate patch

fs/fs_zenfs.cc
yhr: The code is already doing that, and it depends on how much space is available on the disk

fs/fs_zenfs.cc
```diff
@@ -253,11 +256,65 @@ ZenFS::~ZenFS() {
...
+      if (migrate_zone_ids.find(ext.zone_start) != migrate_zone_ids.end()) {
+        migrate_exts.push_back(&ext);
+      }
+    }
+
+    if (migrate_exts.size() > 0) {
+      IOStatus s;
+      Info(logger_, "Garbage collecting %d extents \n",
+           (int)migrate_exts.size());
+      s = MigrateExtents(migrate_exts);
```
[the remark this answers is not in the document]
yhr: MigrateExtents should be thread safe, as long as the file is not being written to in parallel (which we can't guarantee at the moment, so we need to ensure that re-opens and migration is mutually exclusive)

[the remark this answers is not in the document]
yhr: There is actually one thing that looks fishy in MigrateExtents, the ReplaceExtents locking looks broken, we need to make sure that readers won't read outdated information

yhr:
I've addressed all comments(except the refactoring things I believe) and increased the gc aggressiveness a bit (slope 2 -> 3). It looks stable and performing nicely, (sysbench and db_bench tests give promising results), but it might need more fine tuning, so let's keep this on a branch until we've finished evaluating this,

Haltz:
Hi guys! A few days ago I tested this garbage collection draft and it stuck. I guess the GC threshold is not good. Considering a case where all zones have valid and invalid data at the same time, ZenFS maybe can not find a empty zone to collect the garbage. Will you take a look at this?
Please tell me if my assumption is wrong! Thanks very much!

aravind-wdc:
> Hi guys! A few days ago I tested this garbage collection draft and it stuck. I guess the GC threshold is not good. Considering a case where all zones have valid and invalid data at the same time, ZenFS maybe can not find a empty zone to collect the garbage. Will you take a look at this? Please tell me if my assumption is wrong! Thanks very much!

Hi Haltz,

Thanks for testing it out.
Could you give more information on what happened with the hang(how to reproduce this? ) ? 
What were you testing on it with (Hardware) ?
What application were you running ? Did you look at the logs at /tmp/znefs_*.log ?

Haltz:
I will give more infomation(maybe a few days later). I was testing this branch on FEMU(a emulator based on QEMU). RocksDB was in the guest virtual machine and using a emulated ZNS SSD as backend storage device.

fs/fs_zenfs.cc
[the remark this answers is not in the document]
yhr: I checked and ReplaceExtents looks ok, read/write locking was introduced as part of #123

fs/fs_zenfs.h
```diff
@@ -452,6 +456,12 @@ class ZenFS : public FileSystemWrapper {
   IOStatus MigrateFileExtents(
       const std::string& fname,
       const std::vector<ZoneExtentSnapshot*>& migrate_exts);
+
+ private:
+  const uint64_t GC_START_LEVEL =
+      20;                      /* Enable GC when < 20% free space available */
+  const uint64_t GC_SLOPE = 3; /* GC agressiveness */
```
MaisenbacherD, line 463: It would be nice to be able to overwrite the default `GC_SLOPE` and `GC_START_LEVEL` though the ZenFS utility in mkfs.

MaisenbacherD (commented):
The GC feature itself looks good to me :)
However, I thing we need to add a commit that fixes the `MigrateFileExtents` as mentioned in the inline comment.

fs/fs_zenfs.cc
```diff
@@ -256,11 +259,67 @@ ZenFS::~ZenFS() {
...
   delete zbd_;
 }
 
+void ZenFS::GCWorker() {
+  while (run_gc_worker_) {
+    usleep(1000 * 1000 * 10);
+
+    uint64_t non_free = zbd_->GetUsedSpace() + zbd_->GetReclaimableSpace();
+    uint64_t free = zbd_->GetFreeSpace();
+    uint64_t free_percent = (100 * free) / (free + non_free);
```
metaspace, line 278: Maybe it would make sense to use floats here? The integer arithmetic is going to be a little bit off.

metaspace (approved):

fs/fs_zenfs.cc
```diff
@@ -258,11 +259,67 @@ ZenFS::~ZenFS() {
...
+          migrate_zones_start.end()) {
+        migrate_exts.push_back(&ext);
+      }
+    }
+
+    if (migrate_exts.size() > 0) {
+      IOStatus s;
+      Info(logger_, "Garbage collecting %d extents \n",
+           (int)migrate_exts.size());
+      s = MigrateExtents(migrate_exts);
```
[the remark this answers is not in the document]
MaisenbacherD: As we discussed, `ResetUnusedIOZones` does not clean zones which where not migrated by not acquiring the WR lock beforehand.

yhr:
I have some follow-up patches to this pull request(including migration bug fixes), but will merge this and then submit the rest as separate PRs.

PR #186, opened 2022-04-11
zbd: Fix partial I/O handling in Append and DirectRead

### LINKED ISSUE #181: ZonedBlockDevice::DirectRead and Zone::Append don't handle partial I/O correctly
In Zone::Append, the pwrite should use left instead of size as the buffer size.
In ZonedBlockDevice::DirectRead, ret is incremented with the bytes read on every read, but the read is always for the full buffer at the same offset, so, e.g., two reads of the first half a buffer will be viewed as a successful read.

jsvaerke:
In Zone::Append, the pwrite is changed to use left instead of size as the buffer size.
In ZonedBlockDevice::DirectRead, logic has been changed to the same approach as in other
places in the code, where the number of bytes left is tracked and buffer pointers and
offsets are adjusted with the bytes already read.

Signed-off-by: Jorgen S. Hansen <jorgen.hansen@wdc.com>

Closes #181

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -150,7 +150,7 @@ IOStatus Zone::Append(char *data, uint32_t size) {
   assert((size % zbd_->GetBlockSize()) == 0);
 
   while (left) {
-    ret = pwrite(fd, ptr, size, wp_);
+    ret = pwrite(fd, ptr, left, wp_);
     if (ret < 0) {
       return IOStatus::IOError(strerror(errno));
     }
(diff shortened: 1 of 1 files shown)


jsvaerke:
I'm not sure that the DirectRead part is covered by the smoke-test, and I'll be happy to run any additional tests.

MaisenbacherD (approved):
Nice catch! 
Looks good to me.

yhr:
DirectRead is only called during migration, which is not tested by the current tests. I tested this branch on my garbage collection branch and it checked out.

yhr (approved):
Nice, thanks!

PR #187, opened 2022-05-06
util: Allow restore of single files in to ZenFS.

aravind-wdc:
zenfs util allows only directories to be restored, but backup
command allows single file backups, so this makes the
restore behavior consistent with backup command regarding
single file backup/restore.

### DIFF
--- a/util/zenfs.cc
+++ b/util/zenfs.cc
@@ -670,6 +670,9 @@ int zenfs_tool_remove_directory() {
 int zenfs_tool_restore() {
   Status status;
   IOStatus io_status;
+  IOOptions opts;
+  IODebugContext dbg;
+  bool is_dir;
 
   if (FLAGS_path.empty()) {
     fprintf(stderr, "Error: Specify --path to be restored.\n");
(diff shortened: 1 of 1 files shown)


util/zenfs.cc
```diff
@@ -670,15 +670,24 @@ int zenfs_tool_remove_directory() {
...
 
   AddDirSeparatorAtEnd(FLAGS_restore_path);
-  AddDirSeparatorAtEnd(FLAGS_path);
-  ReadWriteLifeTimeHints();
+  std::filesystem::path fpath(FLAGS_path);
+  FLAGS_path = fpath.lexically_normal().string();
+  FileSystem *f_fs = FileSystem::Default().get();
+  status = f_fs->IsDirectory(FLAGS_path, opts, &is_dir, &dbg);
+  if (!status.ok()) {
+    fprintf(stderr, "Failed with %s\n", status.ToString().c_str());
```
yhr, line 688: Please print what failed,align with the other IsDirectory error print

yhr (changes requested):

MaisenbacherD (approved):

metaspace (approved):
Looks good to me!

util/zenfs.cc
aravind-wdc: This does print what failed. Here is the error print in case of failure here:
"Failed with IO error: While open for IsDirectory(): /tmp/bkp1/chennai/: Not a directory "

Anyhow, I have aligned this to the other error print as below:
fprintf(stderr, "IsDirectory failed, error: %s\n",<snip>);

PR #192, opened 2022-05-16
fs: add a Path class and remove c++17 dependency

royguo:
- Add a new Path class to replace the `std::filesystem::path` dependency
- Simplify & safety some of the path operations

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -271,9 +271,8 @@ IOStatus ZenFS::Repair() {
   return IOStatus::OK();
 }
 
-std::string ZenFS::FormatPathLexically(std::filesystem::path filepath) {
-  std::filesystem::path ret = "/" / filepath.lexically_normal();
-  return ret.string();
+std::string ZenFS::FormatPathLexically(std::string filepath) {
+  return Path("/" + filepath).Normalize();
 }
 
 void ZenFS::LogFiles() {
@@ -638,19 +637,12 @@ IOStatus ZenFS::ReopenWritableFile(const std::string& filename,
 void ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
                                    bool include_grandchildren,
                                    std::vector<std::string>* result) {
-  auto path_as_string_with_separator_at_end =
-      [](std::filesystem::path const& path) {
-        auto with_sep = path / "";
-        return with_sep.lexically_normal().string();
-      };
-
   auto string_starts_with = [](std::string const& string,
                                std::string const& needle) {
     return string.rfind(needle, 0) == 0;
   };
 
-  std::string dir_with_terminating_seperator =
-      path_as_string_with_separator_at_end(std::filesystem::path(dir));
+  std::string dir_with_terminating_seperator = Path(dir + "/").Normalize();
 
   auto relative_child_path =
       [&dir_with_terminating_seperator](std::string const& full_path) {
@@ -658,16 +650,15 @@ void ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
       };
 
   for (auto const& it : files_) {
-    std::filesystem::path file_path(it.first);
-    assert(file_path.has_filename());
+    Path file_path(it.first);
+    assert(!file_path.Filename().empty());
 
-    std::string file_dir =
-        path_as_string_with_separator_at_end(file_path.parent_path());
+    std::string file_dir = file_path.Parent();
 
     if (string_starts_with(file_dir, dir_with_terminating_seperator)) {
       if (include_grandchildren ||
           file_dir.length() == dir_with_terminating_seperator.length()) {
-        result->push_back(relative_child_path(file_path));
+        result->push_back(relative_child_path(file_path.Normalize()));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


metaspace:
I strongly object to replicating library code inside ZenFS. #193 is a better solution since it picks the functionality from an existing well reviewed library. If using boost is not acceptable for TerakDB, perhaps TerakDB could maintain the replicated `Path` class in their own tree and we could include that instead of boost if building for TerakDB?

royguo:
@MatiasBjorling @metaspace 
We will upgrade to C++17 if it helps us get much better performance (e.g, >=20%), or resolves some existing bugs.....But for ZenFS, seems the only reason is to use std::filesystem::path (which I think is not worth upgrading to C++17)

I agree with @MatiasBjorling because there are a bunch of existing open-source projects that are still using old versions of RocksDB(e.g. Ceph), we cannot assume that to have C++17 or boost library (unless this `std::filesystem::path` is not replaceable....)

metaspace:
> Quick question. ZenFS is a common library for user-space applications, can it really require that applications uses c++17? Could it make sense that it is only required in case that the app already makes use of c++17 functionality?

@MatiasBjorling As I stated elsewhere c++17 was stable in LLVM and GCC 4 years ago. It's not really bleeding edge stuff here.

We moved to c++17 because RocsDB _requires_ c++17. There is no way to even add compiler flags when building in RocksDB that forbids c++17 features.

The reason this is causing problems for us is that ZenFS is not a proper library. It is a plugin distributed as a source that you compile together with your plugin host by adding the sources to your plugin host source tree. We are relying on whatever build system the plugin host provides, and we must be compiled with the same compiler as the plugin host (because c++ has no stable ABI). If we were to distribute ZenFS as a proper library with a C interface, none of this would be an issue.

If we were to keep ZenFS at c++11, we could add a CI step that builds ZenFS in TerakDB, to catch use of features >= c++17. I don't like that solution, because c++17 is a quality of life improvement over c++11. I think TerakDB should just update to c++17 or reimplement whatever c++17 or even boost library features that they need inside TerakDB code base.

If we decide to keep ZenFS at ancient c++11 standard, and if we can have no dependency on boost, we are going to end up reimplementing library code inside ZenFS. It is more code to review and maintain, more places for bugs. It is work that we do not actually have to deal with if we just go to c++17.

MatiasBjorling:
> > Quick question. ZenFS is a common library for user-space applications, can it really require that applications uses c++17? Could it make sense that it is only required in case that the app already makes use of c++17 functionality?
> 
> @MatiasBjorling As I stated elsewhere c++17 was stable in LLVM and GCC 4 years ago. It's not really bleeding edge stuff here.

You're right it is not bleeding edge, I am not disputing that c++17 as a standard is fine and ready to be used for production. 

> 
> We moved to c++17 because RocsDB _requires_ c++17. There is no way to even add compiler flags when building in RocksDB that forbids c++17 features.

The code could distinguish using the #ifdef CPP11 / #ifdef CPP17 constants. 

> 
> The reason this is causing problems for us is that ZenFS is not a proper library. It is a plugin distributed as a source that you compile together with your plugin host by adding the sources to your plugin host source tree. We are relying on whatever build system the plugin host provides, and we must be compiled with the same compiler as the plugin host (because c++ has no stable ABI). If we were to distribute ZenFS as a proper library with a C interface, none of this would be an issue.

Ok, that might be something to work towards.

> 
> If we were to keep ZenFS at c++11, we could add a CI step that builds ZenFS in TerakDB, to catch use of features >= c++17. I don't like that solution, because c++17 is a quality of life improvement over c++11. I think TerakDB should just update to c++17 or reimplement whatever c++17 or even boost library features that they need inside TerakDB code base.

While I always strive to improve quality of life, we also have to consider that other types of quality of life can come from staying with existing frameworks and inherited stability.

For example, new releases of RocksDB 6.29 are contiguously made after the release of 7.0, therefore, should the CI therefore continue to test 6.29.x? 

I am not proposing to avoid c++17 features, I propose to use it when compiled with projects where a specific standard is supported (newer standard). For example, if RocksDB 7.x compiled ZenFS, and it requires c++17 features at the API level, then zenfs code would need to have that ifdef'ed appropriately. For RocksDB 6.9.x and TerarkDB, c++17 features wouldn't be used, and therefore would work anyway.

> 
> If we decide to keep ZenFS at ancient c++11 standard, and if we can have no dependency on boost, we are going to end up reimplementing library code inside ZenFS. It is more code to review and maintain, more places for bugs. It is work that we do not actually have to deal with if we just go to c++17.

I believe that right now, zenfs code would be required to only use features available in the lowest common denominator c++ standard - i.e., TerarkDB and RocksDB 6.29 adhere to the language features defined up to c++11, whereas RocksDB 7+ can make sure of c++17 specific features.

MatiasBjorling:
@royguo I see that TerarkDB already has a dependency on boost::filesystem - So that would be okay to use?

yhr:
> @royguo I see that TerarkDB already has a dependency on boost::filesystem - So that would be okay to use?

@royguo told me that they have dropped the boost dependency.

@metaspace As we ifdef the implementaion of the breaking API change for plugin registration, we are still compatible with rocksdb 6, no?

metaspace:
> I guess we could continue to support RocksDB 6 in ZenFS 2, but that is not something we are currently discussing. Perhaps we should?

It seems I am wrong. We actually do support both RocksDB 6 and 7 APIs in ZenFS 2. The only thing that prevents us from building in vanilla RocksDB 6 is the C++17 features we use.

yhr:
Fixed in #193

Issue #191, opened 2022-05-16
Remove depenecy on c++17 for terarkdb


yhr:
We are using std::filesystem introduced in c++, which is fine for upstream rocksdb which has switched to the new compiler, but terarkdb's users are still on c++11 for now so we need to provide a separate solution for terkarkb.

If we can provide a good solution(removing the need for std::filesystem) that works for both c++11 and c++17, lets move on to that, otherwise we'll need to isolate the use and provide a separate implementation for terarkdb using ifdefs.

MatiasBjorling:
Is there any feedback from the RocksDB folks if they'll begin to use more c++17 features in the upcoming releases? or would it make sense to extract the features in the filesystem c++17 API and not depend on c++17 features for now?

metaspace:
RocksDB are **_requiring_** C++17 now. They have merged a few things that require c++17, see [here](https://github.com/facebook/rocksdb/pulls?q=+%22c%2B%2B+17%22).

I would argue that we  keep targeting c++17 and fall back to boost in case that is not available. We should be able to check the value of ` __cplusplus` to determine what c++ version we are being compiled with.

If TerakDB does not want a dependency on boost, we can implement the stuff by hand. I think it makes more sense to include reviewed code from boost than rolling our own stuff though.

royguo:
For important online services, we don't dare to upgrade the compiler unless we find some critical bugs in older versions. AFAIK, lots of companies are still using RocksDB 5.x. This is because any upgrade could possibly lead to new bugs (from now code), so we prefer STABLE rather than FAST....

metaspace:
> For important online services, we don't dare to upgrade the compiler unless we find some critical bugs in older versions. AFAIK, lots of companies are still using RocksDB 5.x. This is because any upgrade could possibly lead to new bugs (from now code), so we prefer STABLE rather than FAST....

ZenFS 2.x supports RocksDB >= 7, and RocksDB 7 requires a C++17 compiler. The ZenFS 1.x releases are compatible with RocksDB 6 and compiles with a C++11 compiler.

C++17 support was complete in gcc 8.1.0 which was released May 2018 and clang 7.0.0 that was released in September 2018. That is 4 years ago, not really bleeding edge stuff :shrug:

yhr:
Fix merged

PR #193, opened 2022-05-16
fs, util: Use boost libraries for C++ std < C++17

MaisenbacherD:
C++17 was mandatory, since we were using std::filesystem(::path) and
RocksDB also moved on to C++17.

With this commit boost::filesystem replaces std::filesystem if
the C++ std < C++17 when compiling.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -271,8 +271,8 @@ IOStatus ZenFS::Repair() {
   return IOStatus::OK();
 }
 
-std::string ZenFS::FormatPathLexically(std::filesystem::path filepath) {
-  std::filesystem::path ret = "/" / filepath.lexically_normal();
+std::string ZenFS::FormatPathLexically(fs::path filepath) {
+  fs::path ret = fs::path("/") / filepath.lexically_normal();
   return ret.string();
 }
 
@@ -638,11 +638,10 @@ IOStatus ZenFS::ReopenWritableFile(const std::string& filename,
 void ZenFS::GetZenFSChildrenNoLock(const std::string& dir,
                                    bool include_grandchildren,
                                    std::vector<std::string>* result) {
-  auto path_as_string_with_separator_at_end =
-      [](std::filesystem::path const& path) {
-        auto with_sep = path / "";
-        return with_sep.lexically_normal().string();
-      };
+  auto path_as_string_with_separator_at_end = [](fs::path const& path) {
+    fs::path with_sep = path / fs::path("");
+    return with_sep.lexically_normal().string();
+  };
 
   auto string_starts_with = [](std::string const& string,
                                std::string const& needle) {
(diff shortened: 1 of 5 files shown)


MaisenbacherD:
Adresses https://github.com/westerndigitalcorporation/zenfs/issues/191

fs/fs_zenfs.h
```diff
@@ -6,7 +6,14 @@
 
 #pragma once
 
+#if __cplusplus < 201703L
+#include <boost/filesystem.hpp>
```
royguo, line 10: Maybe we should also check whether boost exists, if not, we can provide another solution for developers.....

metaspace: Lets make sure that we find a solution that also works for TerakDB. Not sure they want boost. Maybe we can detect building under TerakDB and then include TerakDB path logic somehow.

metaspace (changes requested):

fs/fs_zenfs.cc
```diff
@@ -271,8 +271,13 @@ IOStatus ZenFS::Repair() {
...
 
-std::string ZenFS::FormatPathLexically(std::filesystem::path filepath) {
-  std::filesystem::path ret = "/" / filepath.lexically_normal();
+std::string ZenFS::FormatPathLexically(fs::path filepath) {
+  fs::path ret;
+  if (filepath.string().rfind("/", 0) == 0) {
+    ret = filepath.lexically_normal();
+  } else {
+    ret = fs::path("/") / filepath.lexically_normal();
+  }
```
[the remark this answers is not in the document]
MaisenbacherD: The `boost::filesystem::path` implementation of the `operator/ ` differs for the implementation in `std::filesystem::path`.
In `std::filesystem::path`:
`std::filesystem::path("/") / std::filesystem::path("/root/test")` becomes "/root/test"
whereas in `boost::filesystem::path`:
`boost::filesystem::path("/") / boost::filesystem::path("/root/test")` becomes "//root/test" and stays like that even when lexically_normal is applied afterwards.

fs/fs_zenfs.h
MaisenbacherD: I will provide a draft of an implementation that imports the and namespaces the `Path` class from a different file (which can eventually live in a TerakDB).

MaisenbacherD: @royguo I am considering to namespace the `experimental/filesystem` library for c++11. https://en.cppreference.com/w/cpp/header/experimental/filesystem
Would that be something that works out for you?
Also what compiler are you using?

MaisenbacherD: We can't use experimental/filesystem because it does not come with the `lexically_nomal` function. So we are back to the `Path` class in a different file.

MaisenbacherD:
We internally discussed the conversation of https://github.com/westerndigitalcorporation/zenfs/pull/192 and concluded, that ZenFS should support c++11 and c++17 out of the box without any dependencies. Therefore, we added the `filesystem_utility.h` implementation of @royguo to this repo, and make use out of it in case ZenFS is being compiled with c++11. With c++17 we use the `std::filesystem` implementation.

yhr:
@MaisenbacherD : The code looks like it's ready for review, but the commit message still says "use boost", could you update?

MaisenbacherD:
@yhr I adjusted the commit message :)

metaspace (approved):

PR #194, opened 2022-05-17
zbd: fix the stats of empty zones

royguo:
A quick fix for the stats log (this bug rarely happen)

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -428,12 +428,13 @@ void ZonedBlockDevice::LogGarbageInfo() {
   // the result to be precise.
   int zone_gc_stat[12] = {0};
   for (auto z : io_zones) {
-    if (z->IsEmpty()) {
-      zone_gc_stat[0]++;
+    if (z->IsBusy() || !z->Acquire()) {
       continue;
     }
 
-    if (!z->Acquire()) {
+    if (z->IsEmpty()) {
+      zone_gc_stat[0]++;
+      z->Release();
       continue;
     }
 


fs/zbd_zenfs.cc
```diff
@@ -437,6 +437,12 @@ void ZonedBlockDevice::LogGarbageInfo() {
       continue;
     }
 
+    // Need double check in case the zone become empty before we acquire
```
yhr, line 440: The zone needs to be released before we continue. We should also be able to remove the empty check before Aquire(). One check should be enough?

yhr (changes requested):

fs/zbd_zenfs.cc
royguo: Yes, you are right, the first check could be removed.

I kept it there because I wanted to avoid zone acquiring as much as possible, but I am not sure how much optimization we can get.....If you think this optimize doesn't matter at all, I could remove it anyway

fs/zbd_zenfs.cc
```diff
@@ -428,12 +428,13 @@ void ZonedBlockDevice::LogGarbageInfo() {
   // the result to be precise.
   int zone_gc_stat[12] = {0};
   for (auto z : io_zones) {
-    if (z->IsEmpty()) {
-      zone_gc_stat[0]++;
+    if (z->IsBusy() || !z->Acquire()) {
```
yhr, line 431: Please drop the IsBusy call, it is checked by Acquire() anyway.

yhr (changes requested):

fs/zbd_zenfs.cc
metaspace: Also the code will not work as intended, because `IsBusy()` uses relaxed memory ordering. You have to obtain the lock with `Acquire()` before using `IsBusy` if you want proper memory ordering, otherwise there are no guarantees about ordering of the writes and reads of the busy flag.

yhr:
pulled in after removing the isbusy call in #219

PR #198, opened 2022-05-25
fs: dont do memcopy with overlapping memory ranges

yhr:
When preparing buffered writes for the next io,
we copy data within the buffer and that may lead
to overlapping memory ranges that is dissallowed
by memcopy. Use memmove instead to fix this.


percona-ysorokin (commented):
This patch fixes the problem reported by ASan.

PR #206, opened 2022-06-16
Test runner improvements

yhr:
Add some hooks to allow for arbitrary pre/post commands to be run for each tests and for copying rocksdb LOG files (useful for checking write stats in read-while-writing tests that does not include the write IOPS for example)

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -62,6 +62,13 @@ do
     echo "$(tput setaf 1)FAILED$(tput sgr 0)"
     FAILED_TESTS=$((FAILED_TESTS+1))
   fi
+
+  if [ -v LOG_FILE_PATH ]; then
+    set -e
+    cp $LOG_FILE_PATH "$RESULT_PATH/${TEST/.sh/.LOG}"
+    set +e
+  fi
+
 done
 
 echo ""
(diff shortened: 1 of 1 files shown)


metaspace (approved):
:+1:

tests/run.sh
```diff
@@ -62,6 +62,13 @@ do
     echo "$(tput setaf 1)FAILED$(tput sgr 0)"
     FAILED_TESTS=$((FAILED_TESTS+1))
   fi
+
+  if [ -v LOG_FILE_PATH ]; then
+    set -e
```
jsvaerke, line 73: Do you want to fail the test run if the copy fails? Also, -e is already active here, so you are switching to +e

tests/run.sh
```diff
@@ -62,6 +68,19 @@ do
...
+    cp $LOG_FILE_PATH "$RESULT_PATH/${TEST/.sh/.LOG}"
+    set +e
+  fi
```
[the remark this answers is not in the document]
yhr: I considered that, but would like the output to be separated when running both write amp and log file monitoring per test

tests/run.sh
yhr: will fix

jsvaerke (approved):

PR #209, opened 2022-06-17
db_stress verification

MaisenbacherD:
This PR fixes a verification error that occurred when executing db_stress (RocksDB v6.29.5):

```
./db_stress --ops_per_thread=1000 --reopen=5 --db=dbname --fs-uri=zenfs://dev:nvmeXnX --read_fault_one_in=0
```

<del>When boost::filesystem is enabled instead of filesystem_utility.h the "fs: Do not prepend path with '/' in `FormatPathLexically`" commit is necessary otherwise executing db_stress will end in the following error:
`open error: IO error: RenameLink: Failed to find the linked file`</del>

<del>If filesystem_utility.h is enabled the "fs: Do not prepend path with '/' in `FormatPathLexically`" commit is not strictly necessary.</del>

PR #214, opened 2022-07-17
fs: add directory separator for ToAuxPath to avoid some error

dongbula:
if aux path is not end with "/",  some error may happen, this pr can fix it


MaisenbacherD:
Hi @dongbula, 
Can you provide an example with the error message that you are fixing with this commit?

Are you using c++17 (and thus the `std::filesystem::path`)? 
Or are you using c++11 which utilises [`filesystem_utility.h`](https://github.com/westerndigitalcorporation/zenfs/blob/master/fs/filesystem_utility.h)?

If you are using c++11 this PR might fix your issue: https://github.com/westerndigitalcorporation/zenfs/pull/209

PR #215, opened 2022-07-19
build: make zenfs work with cmake build

jsvaerke:
rocksdb has a guideline for how to integrate with the cmake
build system. Simply provide plugin specific variables with
.cc files and the library requirements.

We run generate_version.sh at cmake time - this should be
ok since any change in the git repo could affect cmake,
and as such cmake should be rerun anyways. Also, with
the current cmake infrastructure in rocksdb, it doesn't
seem to be possible to run generate_version.sh at build
time prior to processing the zenfs source files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,17 @@
+cmake_minimum_required(VERSION 3.4)
+
+set(zenfs_SOURCES "fs/fs_zenfs.cc" "fs/zbd_zenfs.cc" "fs/io_zenfs.cc" PARENT_SCOPE)
+set(zenfs_HEADERS "fs/fs_zenfs.h" "fs/zbd_zenfs.h" "fs/io_zenfs.h" "fs/version.h" "fs/metrics.h"
+    "fs/snapshot.h" "fs/filesystem_utility.h" PARENT_SCOPE)
+set(zenfs_LIBS "zbd" PARENT_SCOPE)
+set(zenfs_CMAKE_EXE_LINKER_FLAGS "-u zenfs_filesystems_reg" PARENT_SCOPE)
+
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(ZBD REQUIRED libzbd>=1.5.0)
+
+execute_process(WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                RESULT_VARIABLE GENVER_RESULT
+                COMMAND ./generate-version.sh)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.4)
+
+set(zenfs_SOURCES "fs/fs_zenfs.cc" "fs/zbd_zenfs.cc" "fs/io_zenfs.cc" PARENT_SCOPE)
```
MaisenbacherD, line 3: I think we should also set the `zenfs_HEADERS` accordingly, so that ZenFS is rebuild when changes in the header files are made.

CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+set(zenfs_LIBS "zbd" PARENT_SCOPE)
+set(zenfs_CMAKE_EXE_LINKER_FLAGS "-u zenfs_filesystems_reg" PARENT_SCOPE)
+
+execute_process(WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                RESULT_VARIABLE GENVER_RESULT
+                COMMAND ./generate-version.sh)
+if(GENVER_RESULT AND NOT GENVER_RESULT EQUAL 0)
+  message(FATAL_ERROR "Generating ZenFS version failed")
+endif()
```
MaisenbacherD, line 12: Is there also a way to add a check for the correct libzbd version like here? :

https://github.com/westerndigitalcorporation/zenfs/blob/master/zenfs.mk#L12

MaisenbacherD (commented):
Looks like a nice solution!
Had some comments :)

CMakeLists.txt
jsvaerke: Thanks for pointing that out. zenfs_HEADERS isn't used when building - as I understand it, cmake is able to figure that out on it's own. However, the zenfs_HEADERS is used during install to specify the headers that need to be installed with the package, so we need to specify them.

CMakeLists.txt
jsvaerke: I did not find a way to do that. I'll investigate it a bit more.

jsvaerke: It is possible to add the version check when running cmake, so I've added that. Updating the rocksdb.pc file won't be part of this change.

yhr (approved):
looks good to me

MaisenbacherD:
Looks good to me :)

metaspace (approved):

MaisenbacherD (approved):

PR #225, opened 2022-09-21
fs: Provide shim for NewZenFS calls that don't know about zonefs

jsvaerke:
External callers of NewZenFS may not be aware of the zonefs support, so make a shim that converts from the "old" signature to the new one.


jsvaerke:
OK, needs more work.

yhr (approved):
looks all-right to me!

PR #226, opened 2022-10-03
cmake: Make it possible to skip libzbd check

jsvaerke:
Introduce an environment variable check that makes it possible to skip the libzbd check in cases where libzbd isn't provided througha package.


yhr (approved):
consensus is that this is passible

PR #231, opened 2022-10-06
util: Fix occurances where back() might be called on an empty string

MaisenbacherD:
Replace the pattern where back() might be called on an empty string with the AddDirSeparatorAtEnd fuction which was fixed in a previours commit:

`util: Fix AddDirSeparatorAtEnd where back() can not be called on an empty string`


aravind-wdc (approved):
LGTM

yhr (approved):

PR #237, opened 2022-10-13
zenfs: Do not Repair() when the mount is read-only mode.

aravind-wdc:
Observed that when heavy workloads are running(like sysbench) on zenfs, commands like "zenfs df/list" would error out with below error log. "Unexexpeted extent length while recovering".

When mounted in read-only mode, the process mounting it is understood to not make any changes to the filesystem. So it is better to do a repair() when the process has the intent to write data/metadata (write mode mount). Also when mounted in read-only mode(by say zenfs list), there could be write happening by RocksDB which can alter the data/metadata state and doing a repair at that time in a read-only mount state is also not warranted.

So do repair when mounted with write permission.

### DIFF
--- a/fs/fs_zenfs.cc
+++ b/fs/fs_zenfs.cc
@@ -1396,8 +1396,10 @@ Status ZenFS::Mount(bool readonly) {
     }
   }
 
-  s = Repair();
-  if (!s.ok()) return s;
+  if (!readonly) {
+    s = Repair();
+    if (!s.ok()) return s;
+  }
 
   if (readonly) {
     Info(logger_, "Mounting READ ONLY");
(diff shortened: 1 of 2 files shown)


yhr (approved):
Looks allright to me

MaisenbacherD (approved):
Looks good to me.
Good catch!

metaspace (approved):

metaspace:
On a related note, we might not want to repair without asking the user first?

yhr:
@metaspace : I think we do want automatic repair (the inline extent format sort of requires us to do it). The repair() function might not be well-named. If we discover metadata inconsistencies we should enable some sort of user-agreed-best effort data-salvage mode, but we don't have any of that yet.

PR #241, opened 2022-10-17
fs: inplement cache invalidation

yhr:
Implement cache invalidation - we do support non-direct reads so tell the page cache to drop all data that the extents map to.


yhr:
Useful when running fio-testing

yhr:
OOps, needs some rework after the rebase i just did

MaisenbacherD (approved):
Looks good to me!

fs/io_zenfs.cc
```diff
@@ -325,6 +325,38 @@ ZoneExtent* ZoneFile::GetExtent(uint64_t file_offset, uint64_t* dev_offset) {
   return NULL;
 }
 
+IOStatus ZoneFile::InvalidateCache(uint64_t pos, uint64_t size) {
```
metaspace, line 328:
```suggestion
IOStatus ZoneFile::InvalidateCache(uint64_t pos, uint64_t length) {
```

fs/io_zenfs.cc
```diff
@@ -325,6 +325,38 @@ ZoneExtent* ZoneFile::GetExtent(uint64_t file_offset, uint64_t* dev_offset) {
...
+    uint64_t dev_offset;
+    ZoneExtent* extent = GetExtent(offset, &dev_offset);
+
+    if (!extent) {
+      s = IOStatus::IOError("Extent not found while invalidating cache");
+      break;
+    }
+
+    uint64_t extent_end = extent->start_ + extent->length_;
+    uint64_t invalidate_size = std::min(left, extent_end - dev_offset);
```
metaspace, line 348:
```suggestion
    uint64_t invalidate_length = std::min(left, extent_end - dev_offset);
```

metaspace (changes requested):
Looks good, but I think we should use `length` rather than size.

PR #250, opened 2022-12-01
Fix Segmentation fault while enable zenfs gc.

ywang-wnlo:
When zone become full, the `wp_` will be negative value, so change to use `max_capacity_`.

### DIFF
--- a/fs/zbd_zenfs.cc
+++ b/fs/zbd_zenfs.cc
@@ -356,8 +356,14 @@ void ZonedBlockDevice::LogGarbageInfo() {
       continue;
     }
 
-    double garbage_rate =
-        double(z->wp_ - z->start_ - z->used_capacity_) / z->max_capacity_;
+    double garbage_rate = 0;
+    if (z->IsFull()) {
+      garbage_rate =
+          double(z->max_capacity_ - z->used_capacity_) / z->max_capacity_;
+    } else {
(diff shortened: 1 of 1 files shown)


fs/zbd_zenfs.cc
```diff
@@ -358,6 +358,10 @@ void ZonedBlockDevice::LogGarbageInfo() {
 
     double garbage_rate =
         double(z->wp_ - z->start_ - z->used_capacity_) / z->max_capacity_;
+    if (z->IsFull()) {
```
yhr, line 361: Please change the code to do the garbage calculation only once (using IsFull() as a condition for how to do it)

yhr (changes requested):
See my comment on the code. Apart from that you need to sign off your patch and push a new revision (git commit --amend --signoff && git push HEAD:<your pull req branch) for us to accept it. Thanks!

ywang-wnlo:
I changed the code, and force-pushed it. Is it OK?

yhr:
> I changed the code, and force-pushed it. Is it OK?

Yes, thanks! Our CI seems to be misbehaving though. I'll run the lint check and merge this manually when i'm back from vacation. Happy holidays @ywang-wnlo

yhr:
Merged via PR #256 . Thanks @ywang-wnlo !

PR #254, opened 2022-12-28
assert(true)

sg20180546:
hello.
I think it is better if assert(true) remove
thanks

### DIFF
--- a/.github/workflows/mtr-test.yaml
+++ b/.github/workflows/mtr-test.yaml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: percona/percona-server
-          ref: Percona-Server-8.0.28-19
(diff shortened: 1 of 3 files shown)


yhr:
@sg20180546 Yes, i think we can drop the unconditional assert. Could you format your commit message to motivate the change and provide your signed-off by line (git commit --amend --signoff?).

The subject of the patch should be something like "fs: drop unconditional assert when running out of metadata zones"
check the git history for good examples of commit messages.

Thanks!

yhr:
@sg20180546 : You'll need to signoff with your real name, not your email account number. (Signed-off-by: SungGin Byeon <email.adress@domain>) . Could you update your patch? Thanks!

yhr:
Will fix this as part of another PR

PR #276, opened 2023-07-28
division better than for loop

sg20180546:


yhr:
@sg20180546 : thanks for the improvement. Could you squash the first two commits, but make sure to add a check that the offset is within the iozone range? Also sign off your patch (git commit --signoff)

yhr:
@sg20180546 : I can fix up the commits, but i'll need your signoff to pull in this change. Could you provide that? Thanks!

Issue #266, opened 2023-04-04
The `lifetime_` of the zone will be lost after remount zenfs


ywang-wnlo:
https://github.com/westerndigitalcorporation/zenfs/blob/b04ca0cb49c03ecf193367e314e9accf21c937f6/fs/zbd_zenfs.cc#L51-L64

The zone is configured with a `lifetime_` of `Env::WLTH_NOT_SET` during initialization, but if the zone already has data(such as remount), it should be configured as the maximum value of the `lifetime_` of the files in it.

https://github.com/westerndigitalcorporation/zenfs/blob/b04ca0cb49c03ecf193367e314e9accf21c937f6/fs/zbd_zenfs.cc#L393-L413

Due to the fact that the `lifetime_` of zones is related to the zone allocation algorithm, which will always return `LIFETIME_DIFF_NOT_GOOD` as the `zone_lifetime` is `Env::WLTH_NOT_SET`

ywang-wnlo:
I tried to fix the problem in [this commit](https://github.com/ywang-wnlo/zenfs-fork/commit/4ce58d9b2da6684552a3394ac584c3203abe3491)

I will submit a pull request if needed

yhr:
@ywang-wnlo , this looks like a good fix, please submit a PR. Pardon the delay in my response :)

PR #283, opened 2023-10-30
fs: rebuild zone lifetime while mount

ywang-wnlo:
https://github.com/westerndigitalcorporation/zenfs/issues/266


aravind-wdc (commented):
Can you please add a little more context to the commit message to reflect the need for this change or use case where this would be beneficial ?
And have you tested the impact of this change during mount on an almost full filesystem(ZenFS) on a large drive? Since we are going through all extents of all files in foreground I think it would add significant delay during mount, when the filesystem is full or almost full.

And also you need to rebase with base branch.

PR #303, opened 2025-03-26
fs: fix compile error on latest RocksDB

yhr:
WritableFile::GetFileSize needs to be imlemented now, so hook it up to make it compilable on latest rocksdb.


jsvaerke (approved):
Looks good

MatiasBjorling (approved):

